PR #69, opened 2018-05-01
Following and modifying Transforms - fixes #2 fixes #3 fixes #4

### LINKED ISSUE #2: Allow a Transform to follow an active Transform from a selection in the scene
The ability for a transform to follow an active transform will make it easier for mapping an alias object onto multiple objects in the scene such as the various SDK setups (e.g. map a proxy object to all available SDK HMDs)

This will help separate concerns with SDK specific game objects.

### LINKED ISSUE #3: Allow multiple Transforms to follow a single Transform
This is the opposite of https://github.com/ExtendRealityLtd/VRTK.Unity.Core/issues/2

and will allow a collection of transforms to track another given transform, which will be useful for moving the 3rd party camera rigs around at the same time

### LINKED ISSUE #4: Modify a Transform with constraints (e.g. offset) with a potential transition time
The ability to modify a Transform's position, rotation, scale with a given offset Transform and to be able to perform the transition over a given duration.

This will enable movement of the play area using the headset as an offset position. The transition duration will enable the transform to lerp to it's new orientation.

This will enable teleporting and dash teleporting around a scene.

thestonefox:
The Object Follow script enables the ability to set a source
component to be affected by a collection of target components based
on the injected modifier. The modifier included in this commit is
the TransformSourceFollowActiveTarget which will make the source
follow the first active target found.

The following is done in a generic Moment Processor that removes
the need to pepper scripts with their own Unity game loop methods
such as Update, FixedUpdate, etc.

The Moment Processor allows defined moments to be processed at a
specified point in the Unity game loop, ensuring a single loop
is dealt with and additional controls can be applied such as only
processing the moments if the Component is active in the scene.

### DIFF
--- a/Scripts/Data/Enum/TransformProperties.cs
+++ b/Scripts/Data/Enum/TransformProperties.cs
@@ -1,0 +1,24 @@
+﻿namespace VRTK.Core.Data.Enum
+{
+    using System;
+
+    /// <summary>
+    /// The TransformProperties is an enum collection of properties of a transform.
+    /// </summary>
+    [Flags]
+    public enum TransformProperties
+    {
+        /// <summary>
+        /// The Position of a transform.
+        /// </summary>
+        Position = 1 << 0,
+        /// <summary>
+        /// The Rotation of a transform.
+        /// </summary>
+        Rotation = 1 << 1,
+        /// <summary>
+        /// The Scale of a transform.
+        /// </summary>
+        Scale = 1 << 2
+    }
+}
--- a/Scripts/Data/Type/Vector3State.cs
+++ b/Scripts/Data/Type/Vector3State.cs
@@ -1,0 +1,47 @@
+﻿namespace VRTK.Core.Data.Type
+{
+    using System;
+
+    /// <summary>
+    /// The Vector3State DataType allows a boolean to be set per Vector3 element to provide a state reference.
+    /// </summary>
+    [Serializable]
+    public class Vector3State
+    {
+        /// <summary>
+        /// The X State of the Vector3.
+        /// </summary>
+        public bool xState;
+        /// <summary>
+        /// The Y State of the Vector3.
+        /// </summary>
+        public bool yState;
+        /// <summary>
+        /// The Z State of the Vector3.
+        /// </summary>
+        public bool zState;
+
+        /// <summary>
+        /// Default Vector3State of all false.
+        /// </summary>
+        public static readonly Vector3State False = new Vector3State(false, false, false);
+
+        /// <summary>
+        /// Default Vector3State of all true.
+        /// </summary>
+        public static readonly Vector3State True = new Vector3State(true, true, true);
+
+        /// <summary>
+        /// The Constructor that allows setting the individual states at instantiation.
+        /// </summary>
+        /// <param name="x">The X State.</param>
+        /// <param name="y">The Y State.</param>
+        /// <param name="z">The Z State.</param>
+        public Vector3State(bool x, bool y, bool z)
+        {
+            xState = x;
+            yState = y;
+            zState = z;
+        }
+    }
+}
--- a/Scripts/Process/MomentProcess.cs
+++ b/Scripts/Process/MomentProcess.cs
@@ -1,0 +1,17 @@
+﻿namespace VRTK.Core.Process
+{
+    using UnityEngine;
+    using System;
+
+    /// <summary>
+    /// The MomentProcess is a wrapper for an IProcessable process that has a state to determine when it is to be processed.
+    /// </summary>
+    public sealed class MomentProcess : MonoBehaviour
+    {
+        [SerializeField]
+        [Tooltip("The process to attach to the moment.")]
+        public ProcessContainer process;
+        [Tooltip("Only run the process if the process is on an active GameObject and the component is enabled.")]
+        public bool onlyProcessOnActiveAndEnabled = true;
+    }
+}
--- a/Scripts/Process/MomentProcessor.cs
+++ b/Scripts/Process/MomentProcessor.cs
@@ -1,0 +1,154 @@
+﻿namespace VRTK.Core.Process
+{
+    using UnityEngine;
+
+    /// <summary>
+    /// The MomentProcessor loops through the given moments and executes their Process method on the given Unity game loop moment.
+    /// </summary>
+    public class MomentProcessor : MonoBehaviour
+    {
+        /// <summary>
+        /// The Moment is the point in the Unity game loop when to execute the processes.
+        /// </summary>
+        public enum Moment
+        {
+            /// <summary>
+            /// Never execute the processes.
+            /// </summary>
+            None,
+            /// <summary>
+            /// Execute the processes in the FixedUpdate physics part of the Unity game loop.
+            /// </summary>
+            FixedUpdate,
+            /// <summary>
+            /// Executes the processes in the Update game logic part of the Unity game loop.
+            /// </summary>
+            Update,
+            /// <summary>
+            /// Executes the processes in the LateUpdate game logic part of the Unity game loop.
+            /// </summary>
+            LateUpdate,
+            /// <summary>
+            /// Executes the processes in the camera PreCull scene rendering part of the Unity game loop.
+            /// </summary>
+            PreCull,
+            /// <summary>
+            /// Executes the processes in the camera PreRender scene rendering part of the Unity game loop.
+            /// </summary>
+            PreRender
+        }
+
+        [Tooltip("The moment in which to process the given processes.")]
+        public Moment momentToProcess = Moment.PreRender;
+        [Tooltip("A collection of Moments to process.")]
+        public MomentProcess[] processes = new MomentProcess[0];
+
+        protected Moment subscribedMoment;
+
+        protected virtual void OnEnable()
+        {
+            subscribedMoment = Moment.None;
+            ManageSubscriptions();
+        }
+
+        protected virtual void OnDisable()
+        {
+            UnsubscribeMoment();
+        }
+
+        protected virtual void FixedUpdate()
+        {
+            if (momentToProcess == Moment.FixedUpdate)
+            {
+                Process();
+            }
+        }
+
+        protected virtual void Update()
+        {
+            if (momentToProcess == Moment.Update)
+            {
+                Process();
+            }
+        }
+
+        protected virtual void LateUpdate()
+        {
+            if (momentToProcess == Moment.LateUpdate)
+            {
+                Process();
+            }
+            ManageSubscriptions();
+        }
+
+        protected virtual void OnCameraPreRender(Camera givenCamera)
+        {
+            Process();
+        }
+
+        protected virtual void OnCameraPreCull(Camera givenCamera)
+        {
+            Process();
+        }
+
+        /// <summary>
+        /// The ManageSubscriptions method handles subscribing and unsubscribing to the relevant camera events.
+        /// </summary>
+        protected virtual void ManageSubscriptions()
+        {
+            if (subscribedMoment != momentToProcess)
+            {
+                UnsubscribeMoment();
+                SubscribeMoment();
+            }
+        }
+
+        /// <summary>
+        /// The UnsubscribeMoment method handles unsubscribing to the chosen subscribed moment event.
+        /// </summary>
+        protected virtual void UnsubscribeMoment()
+        {
+            switch (subscribedMoment)
+            {
+                case Moment.PreRender:
+                    Camera.onPreRender -= OnCameraPreRender;
+                    break;
+                case Moment.PreCull:
+                    Camera.onPreCull -= OnCameraPreCull;
+                    break;
+            }
+            subscribedMoment = Moment.None;
+        }
+
+        /// <summary>
+        /// The SubscribeMoment method handles subscribing to the chosen moment to process event.
+        /// </summary>
+        protected virtual void SubscribeMoment()
+        {
+            switch (momentToProcess)
+            {
+                case Moment.PreRender:
+                    Camera.onPreRender += OnCameraPreRender;
+                    break;
+                case Moment.PreCull:
+                    Camera.onPreCull -= OnCameraPreCull;
+                    break;
+            }
+            subscribedMoment = momentToProcess;
+        }
+
+        /// <summary>
+        /// The Process method iterates through the given processes and calls the Process method on each one.
+        /// </summary>
+        protected virtual void Process()
+        {
+            foreach (MomentProcess currentProcess in processes)
+            {
+                if (currentProcess != null && (!currentProcess.onlyProcessOnActiveAndEnabled || currentProcess.isActiveAndEnabled))
+                {
+                    currentProcess.process.Interface.Process();
+                }
+            }
+        }
+    }
+}
--- a/Scripts/Process/ProcessContainer.cs
+++ b/Scripts/Process/ProcessContainer.cs
@@ -1,0 +1,13 @@
+﻿namespace VRTK.Core.Process
+{
+    using System;
+    using VRTK.Core.Utility;
+
+    /// <summary>
+    /// The ProcessContainer is the proxy class used to make a process interface appear in the Unity Inspector.
+    /// </summary>
+    [Serializable]
+    public sealed class ProcessContainer : InterfaceContainer<IProcessable>
+    {
+    }
+}
--- a/Scripts/Tracking/Follow/Modifier/FollowModifier.cs
+++ b/Scripts/Tracking/Follow/Modifier/FollowModifier.cs
@@ -1,0 +1,46 @@
+﻿namespace VRTK.Core.Tracking.Follow.Modifier
+{
+    using UnityEngine;
+
+    /// <summary>
+    /// The FollowModifier provides a way of describing logic to modify a Transform's position, rotation and scale.
+    /// </summary>
+    public abstract class FollowModifier : MonoBehaviour
+    {
+        /// The current source being used in the modifier process.
+        public Transform CachedSource
+        {
+            get;
+            protected set;
+        }
+
+        /// The current target being used in the modifier process.
+        public Transform CachedTarget
+        {
+            get;
+            protected set;
+        }
+
+        /// <summary>
+        /// The ProcessFirstAndActiveOnly method determines whether the FollowModifier should process all targets or just the first active target.
+        /// </summary>
+        /// <returns>Returns `true` if only the first active target should be processed.</returns>
+        public abstract bool ProcessFirstAndActiveOnly();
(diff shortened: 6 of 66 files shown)


Scripts/Process/MomentProcessor.cs
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+
+        protected virtual void LateUpdate()
+        {
+            if (momentToProcess == MomentToProcess.OnLateUpdate)
+            {
+                Process();
+            }
+            ManageSubscriptions();
```
bddckr: I wonder if this needs to be done in multiple places to pick up a change in the `momentToProcess` earlier.

We should probably discuss a generic solution to support variables that show up in the inspector to change at edit-time but that allow changing at runtime, too. I think so far we are doing this "check regularly" approach as well as an "provide setter methods" approach - need to settle on one or even on a new solution.

Scripts/Process/MomentProcessor.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <summary>
+        /// The UnsubscribeMoment method handles unsubscribing to the chosen subscribed moment event.
+        /// </summary>
+        protected virtual void UnsubscribeMoment()
+        {
+            switch (subscribedMoment)
+            {
+                case MomentToProcess.OnPreRender:
+                    Camera.onPreRender -= OnCameraPreRender;
+                    subscribedMoment = MomentToProcess.None;
```
bddckr: It looks like this line is in both cases but should be done regardless of the current value of `subscribedMoment`. I think we should move it to be done after this `switch`.

Scripts/Process/MomentProcessor.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <summary>
+        /// The SubscribeMoment method handles subscribing to the chosen moment to process event.
+        /// </summary>
+        protected virtual void SubscribeMoment()
+        {
+            switch (momentToProcess)
+            {
+                case MomentToProcess.OnPreRender:
+                    Camera.onPreRender += OnCameraPreRender;
+                    subscribedMoment = MomentToProcess.OnPreRender;
```
bddckr: Same duplication. I think we should move it out of the `switch` and instead move it to be done after it. The line should then just read `subscribedMoment = momentToProcess;`.

Scripts/Process/ProcessContainer.cs
```diff
@@ -0,0 +1,51 @@
+﻿namespace VRTK.Core.Process
+{
+    using UnityEngine;
+
+    /// <summary>
+    /// The ProcessContainer acts as a container for a process that can be utilized within a Unity Inspector.
+    /// </summary>
+    public abstract class ProcessContainer
```
bddckr: This is generically usable for **any** interface and as such I think we should change all naming (and docs!) to read `InterfaceContainer` instead. If we do so we need to remember to update the `PropertyDrawer` for it, too.

Scripts/Process/ProcessContainer.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+        }
+        private TInterface _interface;
+    }
+
+    /// <summary>
+    /// The ProcessInterfaceContainer is the proxy class used to make a process interface appear in the Unity Inspector.
+    /// </summary>
+    [System.Serializable]
+    public sealed class ProcessInterfaceContainer : ProcessContainer<IProcessable>
```
bddckr: Once we do the rename to `InterfaceContainer` we should move this into its own file.

Scripts/Process/ProcessContainer.cs
```diff
@@ -0,0 +1,51 @@
...
+                }
+            }
+        }
+        private TInterface _interface;
+    }
+
+    /// <summary>
+    /// The ProcessInterfaceContainer is the proxy class used to make a process interface appear in the Unity Inspector.
+    /// </summary>
+    [System.Serializable]
```
bddckr: Again: Not `using System;` instead?

Scripts/Tracking/Follow/Modifier/FollowModifier.cs
```diff
@@ -0,0 +1,53 @@
...
+    public abstract class FollowModifier : MonoBehaviour
+    {
+        protected Transform cachedSource;
+        protected Transform cachedTarget;
+
+        /// <summary>
+        /// The GetSource method returns the current source being used in the modifier process.
+        /// </summary>
+        /// <returns>A Transform of the current source.</returns>
+        public virtual Transform GetSource()
```
bddckr: We need to discuss this: I'd vote to use `get`-only properties instead of getter methods instead.

Scripts/Tracking/Follow/Modifier/FollowModifier.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            return cachedSource;
+        }
+
+        /// <summary>
+        /// The GetTarget method returns the current targetbeing used in the modifier process.
+        /// </summary>
+        /// <returns>A Transform of the current target.</returns>
+        public virtual Transform GetTarget()
```
bddckr: Again: `get`-only properties instead?

Scripts/Tracking/Follow/Modifier/TransformSourceFollowActiveTarget.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        /// <summary>
+        /// The UpdatePosition method attempts to set the source Transform position to the target Transform position.
+        /// </summary>
+        /// <param name="source">The source to modify.</param>
+        /// <param name="target">The target to utilise in the modification.</param>
+        public override void UpdatePosition(Transform source, Transform target)
+        {
+            if (source != null && target != null)
```
bddckr: Shall we just silently do nothing, or is this something we should be throwing an `ArgumentException` for?

Scripts/Tracking/Follow/Modifier/TransformSourceFollowActiveTarget.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        /// <summary>
+        /// The UpdateRotation method attempts to set the source Transform rotation to the target Transform rotation.
+        /// </summary>
+        /// <param name="source">The source to modify.</param>
+        /// <param name="target">The target to utilise in the modification.</param>
+        public override void UpdateRotation(Transform source, Transform target)
+        {
+            if (source != null && target != null)
```
bddckr: Again: `ArgumentException` instead?

Scripts/Tracking/Follow/ObjectFollow.cs
```diff
@@ -0,0 +1,217 @@
...
+    /// </summary>
+    public class ObjectFollow : SourceTargetProcessor
+    {
+        [Header("Object Follow Settings")]
+
+        [UnityFlag]
+        [Tooltip("The transform properties to follow.")]
+        public TransformProperties follow;
+        [Tooltip("The follow modifier mechanic to apply.")]
+        public FollowModifier followModifier = null;
```
bddckr: I see no need to be explicit about the default value since that's what we'll get without specifying it already.

Scripts/Tracking/Follow/ObjectFollow.cs
```diff
@@ -0,0 +1,217 @@
...
+    /// <summary>
+    /// The ObjectFollow mirrors the transform properties of another object based on the given FollowModifier.
+    /// </summary>
+    public class ObjectFollow : SourceTargetProcessor
+    {
+        [Header("Object Follow Settings")]
+
+        [UnityFlag]
+        [Tooltip("The transform properties to follow.")]
+        public TransformProperties follow;
```
bddckr: Shall we specify a default value for value-types to ensure us changing the enum values in the type itself doesn't change the default of the usages?

Scripts/Process/MomentProcess.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace VRTK.Core.Process
+{
+    using UnityEngine;
+    using System;
+
+    /// <summary>
+    /// The MomentProcess is a wrapper for an IProcessable process that has a state to determine when it is to be processed.
+    /// </summary>
+    [Serializable]
+    public sealed class MomentProcess : MonoBehaviour
```
bddckr: Do we need to inherit from `MonoBehaviour`?
- If we don't: Remove it.
- If we do: Remove the `[Serializable]` above.

Scripts/Tracking/Follow/ObjectFollow.cs
```diff
@@ -0,0 +1,217 @@
...
+        [Tooltip("The follow modifier mechanic to apply.")]
+        public FollowModifier followModifier = null;
+
+        /// <summary>
+        /// The ObjectFollowUnityEvent emits an event with a source Transform and a target Transform.
+        /// </summary>
+        [Serializable]
+        public class ObjectFollowUnityEvent : UnityEvent<Transform, Transform> { };
```
bddckr: Code convention: Empty curly braces on same line or on multiple ones? I'd like to do the same for methods, too.

Scripts/Tracking/Follow/ObjectFollow.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// </summary>
+        public ObjectFollowUnityEvent BeforeProcessed;
+        /// <summary>
+        /// The AfterProcessed event is emitted after all processing is complete.
+        /// </summary>
+        public ObjectFollowUnityEvent AfterProcessed;
+        /// <summary>
+        /// The BeforeTransformUpdated event is emitted before the Transform is updated.
+        /// </summary>
+        public ObjectFollowUnityEvent BeforeTransformUpdated;
```
bddckr: Naming: It's `Processed` everywhere, but these things are `Updated` instead. Shall we rename these and the methods like `UpdatePosition` etc. to use `Process(ed|ing)` instead?

Scripts/Tracking/Follow/ObjectFollow.cs
```diff
@@ -0,0 +1,217 @@
...
+                OnAfterScaleUpdated(sourceTransform, targetTransform);
+            }
+        }
+
+        /// <summary>
+        /// The GetTransformFromComponent method attempts to retrieve the Transform from a given Component.
+        /// </summary>
+        /// <param name="component">The Component to retrieve the Transform from.</param>
+        /// <returns>A Transform if one exists on the given Component.</returns>
+        protected virtual Transform GetTransformFromComponent(Component component)
```
bddckr: This is very generic and doesn't actually need anything from this particular class, nor can I see a need to be able to override this behavior in subclasses. Shall we move it to a `static` helper class instead?

We probably have this need often and sadly Unity's overriding of the `==` and `!=` operators mean we can't use C#'s `?` operator like `component?.transform` inline instead.

Scripts/Data/Enum/TransformProperties.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace VRTK.Core.Data.Enum
+{
+    /// <summary>
+    /// The TransformProperties is an enum collection of properties of a transform.
+    /// </summary>
+    public enum TransformProperties
```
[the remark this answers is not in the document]
thestonefox: 👍

Scripts/Data/Type/Vector3State.cs
```diff
@@ -0,0 +1,57 @@
+﻿namespace VRTK.Core.Data.Type
+{
+    /// <summary>
+    /// The Vector3State DataType allows a boolean to be set per Vector3 element to provide a state reference.
+    /// </summary>
+    [System.Serializable]
```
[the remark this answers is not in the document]
thestonefox: 👍

Scripts/Data/Type/Vector3State.cs
```diff
@@ -0,0 +1,57 @@
...
+        public bool yState;
+        /// <summary>
+        /// The Z State of the Vector3.
+        /// </summary>
+        public bool zState;
+
+        /// <summary>
+        /// Default Vector3State of all false.
+        /// </summary>
+        public static Vector3State False
```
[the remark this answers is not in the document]
thestonefox: +1

Scripts/Data/Type/Vector3State.cs
```diff
@@ -0,0 +1,57 @@
...
+            get
+            {
+                return new Vector3State(false, false, false);
+            }
+        }
+
+        /// <summary>
+        /// Default Vector3State of all true.
+        /// </summary>
+        public static Vector3State True
```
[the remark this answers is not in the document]
thestonefox: 👍

Scripts/Process/MomentProcess.cs
thestonefox: Needs to be `MonoBehaviour` as it has inspector parameters you set in the editor window, so removing `[Serializable]` it is!

Scripts/Process/MomentProcessor.cs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// <summary>
+    /// The MomentProcessor loops through the given moments and executes their Process method on the given Unity game loop moment.
+    /// </summary>
+    public class MomentProcessor : MonoBehaviour
+    {
+        /// <summary>
+        /// The MomentToProcess is the point in the Unity game loop when to execute the processes.
+        /// </summary>
+        public enum MomentToProcess
```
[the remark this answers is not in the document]
thestonefox: 👍

Scripts/Process/MomentProcessor.cs
thestonefox: I don't think we should have setter methods for anything that is a public inspector parameter because you can't change them in the editor at runtime, which means the only thing you can do for things like this is have a polling check.

The reason it only happens in the LateUpdate at the moment is it's not the end of the world if you have to wait for the next frame when changing the param

Scripts/Process/MomentProcessor.cs
thestonefox: 👍

Scripts/Process/MomentProcessor.cs
thestonefox: 👍

Scripts/Process/ProcessContainer.cs
thestonefox: if it's not just for `process`es then what about the `protected Object process` field? should that be called something other than `process`?

Scripts/Process/ProcessContainer.cs
thestonefox: System clashes with UnityEngine.Object, but then if all of these are put in their own files then that shouldn't be an issue

Scripts/Process/ProcessContainer.cs
thestonefox: 👍

Scripts/Tracking/Follow/Modifier/FollowModifier.cs
thestonefox: personally i don't like `readonly` properties i find it confusing that you have a variable that you can't set

like when you have `transform.localScale = something` is fine but `transform.lossyScale = something` doesn't work.

i like variables to be variable :)

Is there something in the msdn guidelines that we could adhere to?

Scripts/Tracking/Follow/Modifier/TransformSourceFollowActiveTarget.cs
thestonefox: I don't think throwing an exception is appropriate. If you don't pass anything then nothing should happen imho

passing two nulls isn't an exception, its just nothing will happen

Scripts/Tracking/Follow/ObjectFollow.cs
thestonefox: you mean do like 

`public TransformProperties follow = TransformProperties.Position | TransformProperties.Rotation | TransformProperties.Scale;`

Scripts/Tracking/Follow/ObjectFollow.cs
thestonefox: 👍

Scripts/Tracking/Follow/ObjectFollow.cs
thestonefox: Hmm tough one, the default formatting is new lines for curly braces, but as this is kinda more like an event declaration it reads nicer on one line.

Scripts/Tracking/Follow/ObjectFollow.cs
thestonefox: hmm, you're not technically processing the position it is actually updating the position.

Process doesn't necessarily mean something will be updated.

And in the point of these events, a `Process` of `Updates` happens and before the `Process` starts you get an event and at the end of the `Process` you get an event, then in between you have `Updates`

Scripts/Tracking/Follow/ObjectFollow.cs
thestonefox: Hmm it is generic but at the moment it's not being used anywhere else other than this file. Plus I'm not sure where else to put it, I don't want to end up having a Utility class that just has static methods in it again.

Scripts/Process/MomentProcessor.cs
bddckr: [This](http://adammechtley.com/2014/06/property-backing-field-drawer/) + [this](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=netframework-4.7.2).

Scripts/Process/ProcessContainer.cs
bddckr: Yup, let's call it `target` perhaps?

Scripts/Process/ProcessContainer.cs
bddckr: Yes on own files. Didn't know about the clash. Was probably never getting that because `Object` and using the "built-in type" `object` are different to prevent a clash.

Scripts/Tracking/Follow/ObjectFollow.cs
bddckr: What about an extension method on `Component`?

`ComponentExtensions.cs`:
```c#
public static class ComponentExtenstions
{
    public static Transform TryGetTransform(this Component component)
    {
        return component == null ? null : component.transform;
    }
}
```

Scripts/Tracking/Follow/ObjectFollow.cs
thestonefox: ok lets go with multi line everywhere then, agreed?

Scripts/Tracking/Follow/ObjectFollow.cs
thestonefox: yeah I like `ComponentExtensions` idea. maybe have a Namespace of `Extension` perhaps in case we want to do additional extensions for other types?

PR #75, opened 2018-05-04
feat(Pointers): pointer concept with straight pointer implementation

thestonefox:
The pointer concept is the ability to designate an area of the scene
with a destination. The Straight Pointer implementation of the pointer
casts a straight beam from an origin point (e.g. controller) and stops
when it collides with an object (or reaches the set maximum length).

A pointer knows when it is activated, deactivated, when it starts
hovering (enters) a new object, when it finishes hovering (exits)
an object and knows what coordinates of the object it is hovering
over. These are all cast as events with a special pointer payload
containing information about the pointer target.

The RayCast emitted by Straight Pointer can be customised by utilising
a PhysicsCast object which is a MonoBehaviour component that allows
edit time configuration of a Physics Cast and can then be injected
into the Pointer script. This can allow optional configuration
such as choosing which layers to ignore and whether to ignore
trigger colliders.

### DIFF
--- a/Prefabs/Pointers.meta
+++ b/Prefabs/Pointers.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: c24914d60df83a847ab3ee50d0ba5916
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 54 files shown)


thestonefox:
*NOTE* ignore this PR it's a dumping ground so don't merge it

I need to rip the bits out into separate PRs

PR #81, opened 2018-05-13
fix(Visual): prevent add from being called twice with same parameters

thestonefox:
The `AddColorOverlay` method will now ignore any subsequent calls of
it with the same parameters within the same frame.

This prevents such things as emitting the event multiple times even
if the method is called multiple times with the same information.


GitCop:
There were the following issues with your Pull Request

 * Commit: 2c58e47b90a1b5e24a756baff085cbffb4f91e6a
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/ExtendRealityLtd/VRTK.Unity.Core

---
This message was auto-generated by https://gitcop.com

PR #82, opened 2018-05-13
Feat/surface locator - fixes #6 - fixes #14

### LINKED ISSUE #6: Ability to determine the nearest surface along a given Vector
To determine the nearest surface will be useful to replicate the `Snap To Nearest Floor` setting but without tying it into any teleport code.

The feature will simply cast a ray in a given direction (e.g. down) and find the nearest surface and report the position of that surface (again with an offset maybe of another object e.g. the headset).

This can then be used to report to move the camera rig down to the nearest surface on teleporting

### LINKED ISSUE #14: Policy lists to enable restrictions on certain actions
Just as in current VRTK, the concept of a policy list to restrict certain actions based on things like the existence of a tag, script or layer on another GameObejct

thestonefox:


Tests/Editor/Tracking/TransformModifyTest.cs
```diff
@@ -38,47 +40,48 @@ public void TearDown()
         [Test]
         public void ModifyPositionNoOffsetInstantTransition()
         {
-            targetObject.transform.position = Vector3.one;
+
```
bddckr: ✂️ Empty line

Scripts/Action/SurfaceChangeAction.cs
```diff
@@ -0,0 +1,67 @@
+﻿namespace VRTK.Core.Action
+{
+    using UnityEngine;
+    using VRTK.Core.Data.Type;
+    using VRTK.Core.Extension;
+
+    /// <summary>
+    /// The SurfaceChangeAction emits an event when all given actions are in their active state.
```
bddckr: Seems like a copy and past mistake: Needs proper explanation what it does.

Scripts/Data/Type/SurfaceData.cs
```diff
@@ -0,0 +1,62 @@
...
+        public RaycastHit PreviousCollisionData
+        {
+            get;
+            protected set;
+        }
+
+        protected RaycastHit _collisionData;
+
+        /// <summary>
+        /// Default Constructor.
```
bddckr: Maybe this should use the same documentation the other constructors get?

`Creates a new SurfaceData with a default origin and direction for the collision search.` would be OK if we apply the other requested change on the other constructor.

Scripts/Data/Type/SurfaceData.cs
```diff
@@ -0,0 +1,62 @@
...
+        public SurfaceData()
+        {
+        }
+
+        /// <summary>
+        /// Creates a new SurfaceData with a default origin and direction for the collision search.
+        /// </summary>
+        /// <param name="defaultOrigin"></param>
+        /// <param name="defaultDirection"></param>
+        public SurfaceData(Vector3 defaultOrigin, Vector3 defaultDirection)
```
bddckr: Why are we using a prefix of `default` for the parameter names? I'd like to just use the names of the fields/properties (use `this.origin = origin` inside the constructor block).

Scripts/Data/Type/TransformData.cs
```diff
@@ -0,0 +1,69 @@
...
+        public bool Valid
+        {
+            get
+            {
+                return (transform != null);
+            }
+        }
+
+        /// <summary>
+        /// Default Constructor.
```
bddckr: Maybe this should use the same documentation the other constructors get?
`Creates a new TransformData for an empty Transform.`

PR #83, opened 2018-05-13
feat(Pointer): solid straight pointer - fixes #7

### LINKED ISSUE #7: A pointer that emits a straight visible line with a cursor that reacts to other scene objects
This is the straight pointer but has no connection to teleporters or button input, a simple concept that simply emits a raycast and draws a straight pointer with a cursor that when interacts with objects in the scene will report the items it has entered, is hovering over, has exited and when it is selected.

It will be activatable by a simple method call and no button presses will be coupled to it.

thestonefox:
The pointer concept is the ability to designate an area of the scene
with a destination. The Solid Straight Pointer implementation of the
pointer casts a straight beam from an origin point (e.g. controller)
and stops when it collides with an object (or reaches the set
maximum length).

A pointer knows when it is activated, deactivated, when it starts
hovering (enters) a new object, when it finishes hovering (exits)
an object and knows what coordinates of the object it is hovering
over. These are all cast as events with a special pointer payload
containing information about the pointer target.

### DIFF
--- a/Scripts/Pointer/BasePointer.cs
+++ b/Scripts/Pointer/BasePointer.cs
@@ -1,0 +1,298 @@
+﻿namespace VRTK.Core.Pointer
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using System;
+    using VRTK.Core.Utility;
+    using VRTK.Core.Data.Type;
+
+    /// <summary>
+    /// The PointerData contains information about the current pointer state.
+    /// </summary>
+    public class PointerData : SurfaceData
+    {
+        /// <summary>
+        /// Determines if the pointer is currently in it's activated state.
+        /// </summary>
+        public bool isActive;
+        /// <summary>
+        /// Determines if the pointer is in it's valid state.
+        /// </summary>
+        public bool isValid;
+        /// <summary>
+        /// The duration that the pointer has been hovering over it's current target.
+        /// </summary>
+        public float hoverDuration;
+    }
+
+    /// <summary>
+    /// The BasePointer forms the basis for all pointer types.
+    /// </summary>
+    public abstract class BasePointer : MonoBehaviour
+    {
+        /// <summary>
+        /// The visibility of the pointer element.
+        /// </summary>
+        public enum VisibilityType
+        {
+            /// <summary>
+            /// The element will only be visible when the pointer is in the active state.
+            /// </summary>
+            OnWhenActive,
+            /// <summary>
+            /// The element will always be visible regardless of the pointer state.
+            /// </summary>
+            AlwaysOn,
+            /// <summary>
+            /// The element will never be visible regardless of the pointer state.
+            /// </summary>
+            AlwaysOff
+        }
+
+        [Header("Base Settings")]
+
+        [Tooltip("An optional ExclusionRule to determine valid and invalid targets based on the set rules.")]
+        public ExclusionRule exclusionRule;
+        [Tooltip("An optional custom PhysicsCast object to affect the Cast of the pointer.")]
+        public PhysicsCast physicsCast;
+
+        /// <summary>
+        /// The PointerUnityEvent emits an event with the PointerData payload and the sender object.
+        /// </summary>
+        [Serializable]
+        public class PointerUnityEvent : UnityEvent<PointerData, object>
+        {
+        };
+
+        [Header("Pointer Events")]
+
+        /// <summary>
+        /// The Activated event is emitted when the pointer becomes active.
+        /// </summary>
+        public PointerUnityEvent Activated = new PointerUnityEvent();
+        /// <summary>
+        /// The Deactivated event is emitted when the pointer becomes inactive.
+        /// </summary>
+        public PointerUnityEvent Deactivated = new PointerUnityEvent();
+        /// <summary>
+        /// The Entered event is emitted when the pointer collides with a new target.
+        /// </summary>
+        public PointerUnityEvent Entered = new PointerUnityEvent();
+        /// <summary>
+        /// The Exited event is emitted when the pointer stops colliding with an existing target.
+        /// </summary>
+        public PointerUnityEvent Exited = new PointerUnityEvent();
+        /// <summary>
+        /// The Hovering event is emitted when the pointer changes its hovering position ove an existing target.
+        /// </summary>
+        public PointerUnityEvent Hovering = new PointerUnityEvent();
+        /// <summary>
+        /// 
+        /// </summary>
+        public PointerUnityEvent Selected = new PointerUnityEvent();
+
+        public abstract bool IsVisible();
+
+        /// <summary>
+        /// Reports the active state of the pointer.
+        /// </summary>
+        public bool Active
+        {
+            get;
+            protected set;
+        }
+
+        /// <summary>
+        /// Reports the valid state of the pointer.
+        /// </summary>
+        public bool Valid
+        {
+            get;
+            protected set;
+        }
+
+        /// <summary>
+        /// Reports hover duration of the pointer over the current target.
+        /// </summary>
+        public float HoverDuration
+        {
+            get;
+            protected set;
+        }
+
+        /// <summary>
+        /// Stores the RayCastHit information of the current pointer collision.
+        /// </summary>
+        public RaycastHit CollisionData
+        {
+            get;
+            protected set;
+        }
+
+        /// <summary>
+        /// Stores the Transform information of the cursor state.
+        /// </summary>
+        public Transform CursorData
+        {
+            protected get;
+            set;
+        }
+
+        protected RaycastHit cachedCollision;
+
+        /// <summary>
+        /// The Activate method turns on the pointer.
+        /// </summary>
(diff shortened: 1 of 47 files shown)


Scripts/Pointer/BasePointer.cs
```diff
@@ -0,0 +1,298 @@
...
+            /// <summary>
+            /// The element will never be visible regardless of the pointer state.
+            /// </summary>
+            AlwaysOff
+        }
+
+        [Header("Base Settings")]
+
+        [Tooltip("An optional ExclusionRule to determine valid and invalid targets based on the set rules.")]
+        public ExclusionRule exclusionRule;
```
bddckr: What about renaming this to `validityExclusionRule` to allow classes that inherit from this abstract class to have additional exclusion rules without confusing users because of the naming?

Scripts/Pointer/BasePointer.cs
```diff
@@ -0,0 +1,298 @@
...
+        /// <summary>
+        /// The Exited event is emitted when the pointer stops colliding with an existing target.
+        /// </summary>
+        public PointerUnityEvent Exited = new PointerUnityEvent();
+        /// <summary>
+        /// The Hovering event is emitted when the pointer changes its hovering position ove an existing target.
+        /// </summary>
+        public PointerUnityEvent Hovering = new PointerUnityEvent();
+        /// <summary>
+        ///
```
bddckr: Missing summary.

Scripts/Pointer/BasePointer.cs
```diff
@@ -0,0 +1,298 @@
...
+        }
+
+        /// <summary>
+        /// The IsElementVisible method determines if a given VisibilityType parameter should be considered visible.
+        /// </summary>
+        /// <param name="elementType">The VisibilityType to check for.</param>
+        /// <returns>Returns `true` if the VisibilityType given is considered visible.</returns>
+        public virtual bool IsElementVisible(VisibilityType elementType)
+        {
+            if (!enabled)
```
bddckr: I'm not sure if this should be checking against `enabled` or `isActiveAndEnabled` instead.

Scripts/Pointer/BasePointer.cs
```diff
@@ -0,0 +1,298 @@
...
+                rotationOverride = (givenCursorData != null ? (Quaternion?)givenCursorData.localRotation : null),
+                localScaleOverride = (givenCursorData != null ? (Vector3?)givenCursorData.localScale : null),
+                origin = transform.position,
+                direction = transform.forward,
+                CollisionData = givenCollisionData,
+                isActive = Active,
+                isValid = Valid,
+                hoverDuration = HoverDuration
+            };
+            return e;
```
bddckr: Let's merge the creation and return lines here to prevent the need to have less meaningful variable names like `e`.

Scripts/Pointer/BasePointer.cs
```diff
@@ -0,0 +1,298 @@
...
+        /// 
+        /// </summary>
+        public PointerUnityEvent Selected = new PointerUnityEvent();
+
+        public abstract bool IsVisible();
+
+        /// <summary>
+        /// Reports the active state of the pointer.
+        /// </summary>
+        public bool Active
```
bddckr: Can we perhaps remove this and use `MonoBehaviour`'s `enabled` state instead? I wonder about the confusion it gives having multiple states like these.

Additionally I'd expect the pointer "to turn off" when I disable the component.

Scripts/Pointer/BasePointer.cs
```diff
@@ -0,0 +1,298 @@
...
+
+    /// <summary>
+    /// The BasePointer forms the basis for all pointer types.
+    /// </summary>
+    public abstract class BasePointer : MonoBehaviour
+    {
+        /// <summary>
+        /// The visibility of the pointer element.
+        /// </summary>
+        public enum VisibilityType
```
bddckr: This feels usable in other scenarios, too. Should we move it in the `Type` namespace and adjust the documentation on it to be generic?

Scripts/Pointer/StraightPointer.cs
```diff
@@ -0,0 +1,176 @@
...
+        [Tooltip("The GameObject to represent the pointer's tracer when it is colliding with an invalid object or not colliding at all.")]
+        public GameObject invalidTracer;
+        [Tooltip("The GameObject to represent the pointer's cursor when it is colliding with an invalid object or not colliding at all.")]
+        public GameObject invalidCursor;
+        [Tooltip("Determines when the pointer tracer GameObject should be seen.")]
+        public VisibilityType tracerVisibility = VisibilityType.OnWhenActive;
+        [Tooltip("Determines when the pointer cursor GameObject should be seen.")]
+        public VisibilityType cursorVisibility = VisibilityType.OnWhenActive;
+        [Tooltip("Determines if the pointer should be automatically activated when the script is enabled.")]
+        public bool activateOnEnable;
```
bddckr: See review comment on `BasePointer`: Maybe we can merge "active" and `MonoBehaviour`'s `enabled`?

Scripts/Pointer/StraightPointer.cs
```diff
@@ -0,0 +1,176 @@
...
+                SetElementActive(invalidElement, false);
+            }
+        }
+
+        /// <summary>
+        /// The SetElementActive method determines the active state of the given element.
+        /// </summary>
+        /// <param name="element">The GameObject of which to set it's active state.</param>
+        /// <param name="state">The state to set on the GameObject.</param>
+        protected virtual void SetElementActive(GameObject element, bool state)
```
bddckr: Maybe this should be an extension method on any `GameObject` instead?

Scripts/Pointer/BasePointer.cs
thestonefox: Agreed, but maybe we should move it when there is a need for it to be in multiple places?

bddckr: OK 👍

Scripts/Pointer/BasePointer.cs
thestonefox: `ExclusionRule` isn't an abstract class it's basically the new name for `PolicyList` because it's basically a `Policy` but the word list was dumb

and `PolicyRule` is also dumb because they mean the same thing.

Maybe `ExclusionRule` should be simply called `Policy` and you just check if the policy should exclude the thing you're checking on `Policy.ShouldExclude(thing)`

then in this case you could have the variable called `validityPolicy`

Scripts/Pointer/BasePointer.cs
thestonefox: #pro

Scripts/Pointer/BasePointer.cs
thestonefox: `Active` isn't the same as enabled.

enabled is the component is on

`Active` in the case of a pointer is the pointer has been activated via the `Activate()` method

Scripts/Pointer/BasePointer.cs
bddckr: I think `ExclusionRule` is a good name, though I like `Policy`, too. 🤔

For this comment I only am interested in changing the variable name. I can see subclasses perhaps needing a different "rule" for other functionality and having a generically named `exclusionRule` on the superclass makes that harder to understand.

`exclusionRule`'s documentation already states it's only used "to determine valid and invalid targets".

Scripts/Pointer/BasePointer.cs
thestonefox: I "think" this doesn't use activeandenabled because the actual valid/invalid gameobjects can be disabled.

but then this is actually on the gameobject of the script not the child elements so you may be right. i'll change it and see what breaks

Scripts/Pointer/BasePointer.cs
thestonefox: 👍 Probably look to do this everywhere too (and add to the code conventions docs)

Scripts/Pointer/BasePointer.cs
bddckr: What does a Pointer do when `enabled == false` but `active == true`?

Scripts/Pointer/StraightPointer.cs
thestonefox: this is different again. this is if this is ticked then call the `Activate()` method when the script becomes enabled.

Scripts/Pointer/StraightPointer.cs
thestonefox: I don't think this will work as an extension method because it's checking if the gameobject is null

if the gameobject was null then you wouldn't have extension methods available on it would you? or does an extension method on a gameobject still allow for the object calling the extension to be null? if so then yeah i agree extension is better

bddckr: Yup that works with extension methods and is actually something one has to be aware of when writing extension methods.

Basically all that needs to check out is the type as it's compile-time, the value isn't checked.

bddckr:
Superseded by #86.

PR #94, opened 2018-05-22
feat(Tracking): collision tracking

bddckr:
To track whether a collision between to objects happens Unity offers
so called "Messages" which are defined methods that are not part of
the inheritance chain. This change adds a new component
`CollisionTracker` that allows to forward any of these messages via
UnityEvents. The events differentiate between a trigger and non-
trigger collision and the event data allows retrieving that
information in addition to the collision and the collider.

### DIFF
--- a/Scripts/Tracking/CollisionTracker.cs
+++ b/Scripts/Tracking/CollisionTracker.cs
@@ -1,0 +1,178 @@
+﻿namespace VRTK.Core.Tracking
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using System;
+
+    /// <summary>
+    /// Holds information about a tracked collision.
+    /// </summary>
+    public struct CollisionTrackerData
+    {
+        /// <summary>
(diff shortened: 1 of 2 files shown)


Scripts/Tracking/CollisionTracker.cs
```diff
@@ -0,0 +1,160 @@
...
+                {
+                    isTrigger = true,
+                    collision = null,
+                    collider = collider
+                });
+        }
+
+        protected void OnCollisionEnterEvent(CollisionTrackerData data)
+        {
+            CollisionEnter?.Invoke(data, this);
```
thestonefox: wrap these with `isActiveAndEnabled`

I'm wondering if the `OnTriggerEnter` and the others still work if the component is disabled?

If not then don't bother.

bddckr: Turns out Unity still calls those "message methods". Added the check in!

PR #102, opened 2018-05-23
Feat/teleporters - fixes #98

### LINKED ISSUE #98: Create teleporter prefabs
A collection of prefabs that use the relevant scripts to make a teleporter

thestonefox:


Scripts/Data/Type/CameraList.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace VRTK.Core.Data.Type
+{
+    using UnityEngine;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Provides a collection of scene cameras.
+    /// </summary>
+    public class CameraList : MonoBehaviour
```
bddckr: Change to
```c#
[Serializable]
public class CameraList
```
instead 😄

bddckr: Turns out that's only useful for things you reference on other `MonoBehaviour`s. You can't add `[Serializable]` things as components since they obviously aren't one!

PR #105, opened 2018-05-24
fix(Pointer): ensure exit/deactivate events are called in correct order

thestonefox:
The ObjectPointer `Deactivated` event would always be called even if
the pointer component was disabled. This has been fixed by wrapping
the `Deactivate` method with an `isActiveAndEnabled` but also in the
`OnDisable` a specific `Deactivated` event is emitted otherwise the
event would never fire when the component is disabled at runtime.

Also, the hover duration was being reset to 0 before the exit event
was emitted meaning the exit event didn't include the correct hover
duration.

### DIFF
--- a/Scripts/Pointer/ObjectPointer.cs
+++ b/Scripts/Pointer/ObjectPointer.cs
@@ -228,6 +228,7 @@
         }
 
         protected PointsCastData activePointsCastData;
+        protected PointsCastData previousPointsCastData;
         protected bool? previousVisibility;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Scripts/Pointer/ObjectPointer.cs
```diff
@@ -311,6 +314,7 @@ protected virtual void OnEnable()
         protected virtual void OnDisable()
         {
             Deactivate();
+            Deactivated?.Invoke(HoverTarget, this);
```
bddckr: The `Deactivate` method is doing more than just emitting this event. I wonder if we need anything like the `TryEmitExit` its doing (among the other things) in here, too?

PR #110, opened 2018-06-05
feat(Follow): split follow modifiers into more distinct chunks

thestonefox:
The TransformSourceFollowActiveTarget and TransformTargetsFollowSource
Follow Modifiers have been removed and split down into smaller more
distinct chunks.

Now there is a simple `ModifyTransform` modifier which simply sets the
source transform data to match the target transform data.

The logic of whether to update source to the active target or to update
all targets to the current source has been separated into their own
classes which simply determine that responsibility and take a modifier
as a parameter to choose how to actually modify the transform.

So the logic now is:

 * If source wants to follow the active target by changing transform
   * Apply a ModifyTransform component
   * Apply a SourceFollowActiveTarget component
   * Provide the ModifyTransform component to the appliedModifier
     parameter of the SourceFollowActiveTarget component
 * If the targets want to follow the source by changing transform
   * Apply a ModifyTransform component
   * Apply a TargetFollowSource component
   * Provide the ModifyTransform component to the appliedModifier
     parameter of the TargetFollowSource component

The previous way of determining if the modifier should process the
first active or all targets has now been changed from a bool state
that had an abstract method to receive the information to an enum
that determines how the modifier should be treated. This is more
descriptive and will allow future enhancements if they are required.


Scripts/Tracking/Follow/Modifier/SourceFollowActiveTarget.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <param name="source">The source <see cref="Transform"/> to modify.</param>
+        /// <param name="target">The target <see cref="Transform"/> to utilize in the modification.</param>
+        public override void UpdatePosition(Transform source, Transform target)
+        {
+            if (appliedModifier != null && isActiveAndEnabled)
+            {
+                appliedModifier.UpdatePosition(source, target);
+                CachedSource = appliedModifier.CachedSource;
+                CachedTarget = appliedModifier.CachedTarget;
+            }
```
bddckr: This logic is repeated in each subclass of `FollowModifier`. I wonder if we should change the methods `FollowModifier.Update*` to be non-abstract and implement this, then call a `protected` method that actually implements the "real" `Update*`?

In the superclass `FollowModifier`:
```c#
public virtual void UpdatePosition(Transform source, Transform target)
{
    if (appliedModifier != null && isActiveAndEnabled)
    {
        DoUpdatePosition(source, target);
        CachedSource = appliedModifier.CachedSource;
        CachedTarget = appliedModifier.CachedTarget;
    }
}

protected abstract void UpdatePosition(Transform source, Transform target);
```

In the subclass `SourceFollowActiveTarget`:
```c#
protected override void UpdatePosition(Transform source, Transform target)
{
    appliedModifier.UpdatePosition(source, target);
}
```

In the subclass `TargetsFollowSource`:
```c#
protected override void UpdatePosition(Transform source, Transform target)
{
    appliedModifier.UpdatePosition(target, source);
}
```

I always resort to the `DoMethodName` notation for this. (Plus the same for the other `Update*` methods.)

PR #112, opened 2018-06-09
chore(events): add arguments and sender to raise methods - resolves #108

### LINKED ISSUE #108: Figure out a solution to forward/get the sender of an event that was forwarded with EventForwarder

bddckr:
Components that offer events should always offer methods to raise
them. The method that raises the event always needs to follow the
naming convention and additionally should always take the arguments
for the event, that is the data and sender.

This change applies the necessary changes and additionally
simplifies some of the Action implementations to reduce the number
of methods. Actions run their logic in the event-raising method now
instead of delegating it into additional methods that have no
reusability.

### DIFF
--- a/Scripts/Action/BaseAction.cs
+++ b/Scripts/Action/BaseAction.cs
@@ -1,84 +1,140 @@
 ﻿namespace VRTK.Core.Action
 {
     using UnityEngine;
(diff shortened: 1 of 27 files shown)


bddckr:
Superseded by #120.

PR #117, opened 2018-06-12
chore(Visual): rename camera overlay events

bddckr:
There is no need to name an event after the type it is a member of.
This changes the `CameraColorOverlay` to use the proper event naming
convention that is used everywhere else.


thestonefox:
One thing, rename the scope to `Visual`

as the scope of the change is within the `Visual` namespace

PR #118, opened 2018-06-12
chore(events): add arguments and sender to raise methods

bddckr:
Components that offer events should always offer methods to raise
them. The method that raises the event always needs to follow the
naming convention and additionally should always take the arguments
for the event, that is the data and sender (in that order).

This change applies the necessary changes and additionally renames
the event names of `ObjectPointer` since the raise method
`OnBecameVisible` is a "message method" in Unity already. The
pointer prefabs are both updated for this name change.

### DIFF
--- a/Scripts/Action/BooleanAction.cs
+++ b/Scripts/Action/BooleanAction.cs
@@ -9,7 +9,7 @@
     public class BooleanAction : BaseAction<bool>
     {
         /// <summary>
-        /// Defines the event with the <see cref="bool"/> state and sender <see cref="object"/>.
+        /// Defines the event with the <see cref="bool"/> state and initiator <see cref="object"/>.
         /// </summary>
         [Serializable]
         public class BooleanActionUnityEvent : UnityEvent<bool, object>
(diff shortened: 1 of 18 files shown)


Scripts/Action/BooleanAction.cs
```diff
@@ -30,52 +30,52 @@ public class BooleanActionUnityEvent : UnityEvent<bool, object>
...
             if (HasChanged())
             {
-                OnChanged(value);
+                OnChanged(value, initiator);
             }
         }
 
         /// <inheritdoc />
-        protected override void OnActivated(bool value)
+        protected override void OnActivated(bool value, object sender)
```
thestonefox: change to `initiator` like done in other places

thestonefox: or are you having the `On` methods use `sender` but everything else use `initiator`?

Scripts/Action/CompoundAndAction.cs
```diff
@@ -38,12 +38,12 @@ protected virtual void Update()
 
             if (isValid && !State)
             {
-                OnActivated(true);
+                OnActivated(true, this);
```
thestonefox: shouldn't this be using `initiator`?

thestonefox: or `initiator ?? this`

Scripts/Action/FloatAction.cs
```diff
@@ -73,17 +73,17 @@ protected virtual void EmitEvents()
         {
             if (Activate())
             {
-                OnActivated(Value);
+                OnActivated(Value, this);
```
thestonefox: should be `initiator ?? this` right?

Scripts/Action/ToggleAction.cs
```diff
@@ -6,21 +6,21 @@
...
-        public override void Receive(bool value, object sender = null)
+        public override void Receive(bool value, object initiator = null)
         {
             previousValue = Value;
             Value = value;
 
             if (!State)
             {
-                OnActivated(true);
+                OnActivated(true, initiator);
```
thestonefox: again, should be `initiator ?? this` in case initiator is null then you want to use that script

bddckr:
Superseded by #120.

PR #123, opened 2018-06-17
refactor(Velocity): move common calls into base class

thestonefox:
The Velocity Estimator concrete classes always made a call to
`IsActive` but this logic is consistent across all sub classes so
it has now been moved into the base class and the public methods
are no longer abstract, instead the implementation is a protected
method that is abstract and requires overriding.

### DIFF
--- a/Scripts/Tracking/Velocity/VelocityEstimator.cs
+++ b/Scripts/Tracking/Velocity/VelocityEstimator.cs
@@ -42,18 +42,6 @@
             return (isActiveAndEnabled && source != null && source.gameObject.activeInHierarchy);
         }
 
-        /// <inheritdoc />
-        public override Vector3 GetVelocity()
-        {
-            return GetEstimate(velocitySamples);
-        }
-
-        /// <inheritdoc />
-        public override Vector3 GetAngularVelocity()
-        {
-            return GetEstimate(angularVelocitySamples);
-        }
-
         /// <summary>
         /// Manually enables the collecting of samples for the estimation process.
         /// </summary>
@@ -85,6 +73,11 @@
         /// <returns>Acceleration of the <see cref="source"/>.</returns>
         public virtual Vector3 GetAcceleration()
         {
+            if (!IsActive())
+            {
+                return Vector3.zero;
+            }
+
             Vector3 average = Vector3.zero;
             for (int i = 2 + currentSampleCount - velocitySamples.Length; i < currentSampleCount; i++)
             {
@@ -114,6 +107,18 @@
         protected virtual void LateUpdate()
         {
             ProcessEstimation();
+        }
+
+        /// <inheritdoc />
+        protected override Vector3 DoGetVelocity()
+        {
+            return GetEstimate(velocitySamples);
+        }
+
+        /// <inheritdoc />
+        protected override Vector3 DoGetAngularVelocity()
+        {
+            return GetEstimate(angularVelocitySamples);
         }
 
         /// <summary>
--- a/Scripts/Tracking/Velocity/VelocityTrackerProcessor.cs
+++ b/Scripts/Tracking/Velocity/VelocityTrackerProcessor.cs
@@ -23,10 +23,19 @@
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


Scripts/Tracking/Velocity/VelocityEstimator.cs
```diff
@@ -39,19 +39,7 @@ public class VelocityEstimator : VelocityTracker
...
-        public override Vector3 GetVelocity()
-        {
-            return GetEstimate(velocitySamples);
-        }
-
-        /// <inheritdoc />
-        public override Vector3 GetAngularVelocity()
-        {
-            return GetEstimate(angularVelocitySamples);
+            return (isActiveAndEnabled && source != null && source.gameObject.activeInHierarchy && isActiveAndEnabled);
```
bddckr: Duplicated `isActiveAndEnabled`. Should this just be
```c#
return (isActiveAndEnabled && source != null && source.isActiveAndEnabled);
```
perhaps?

Scripts/Tracking/Velocity/VelocityEstimator.cs
```diff
@@ -116,6 +109,28 @@ protected virtual void LateUpdate()
             ProcessEstimation();
         }
 
+        /// <inheritdoc />
+        protected override Vector3 DoGetVelocity()
+        {
+            if (!IsActive())
```
bddckr: This check is already part of the superclass' checks in `GetVelocity`.

Scripts/Tracking/Velocity/VelocityEstimator.cs
```diff
@@ -116,6 +109,28 @@ protected virtual void LateUpdate()
...
+                return Vector3.zero;
+            }
+
+            return GetEstimate(velocitySamples);
+        }
+
+        /// <inheritdoc />
+        protected override Vector3 DoGetAngularVelocity()
+        {
+            if (!IsActive())
```
bddckr: This check is already part of the superclass' checks in `GetAngularVelocity`.

Scripts/Tracking/Velocity/VelocityTrackerProcessor.cs
```diff
@@ -46,7 +55,7 @@ public override Vector3 GetVelocity()
         /// The reported angular velocity on the first active <see cref="VelocityTracker"/>.
         /// </summary>
         /// <returns>The current angular velocity.</returns>
```
bddckr: This XML docs block should just be `<inheritdoc />`.

Scripts/Tracking/Velocity/VelocityTrackerProcessor.cs
```diff
@@ -22,11 +22,20 @@ public override bool IsActive()
...
+        /// <returns>The current active <see cref="VelocityTracker"/>.</returns>
+        public virtual VelocityTracker GetActiveVelocityTracker()
+        {
+            return (cachedTracker != null && cachedTracker.IsActive() ? cachedTracker : null);
+        }
+
         /// <summary>
         /// The reported velocity on the first active <see cref="VelocityTracker"/>.
         /// </summary>
         /// <returns>The current velocity.</returns>
```
bddckr: This XML docs block should just be `<inheritdoc />`.

Scripts/Tracking/Velocity/VelocityEstimator.cs
thestonefox: source is a transform so has no access to isActiveAndEnabled, but the rest is legit

Scripts/Tracking/Velocity/VelocityTrackerProcessor.cs
thestonefox: The docs are actually different for this method, it's doing something specific

Scripts/Tracking/Velocity/VelocityTrackerProcessor.cs
thestonefox: same here, the docs are specific to the method

PR #124, opened 2018-06-17
chore(Tracking): collision tracker events

bddckr:
The `CollisionTracker` followed Unity's physics message method names
for its event names. This change renames all events to follow the
code conventions and merges the collision and trigger events to only
leave the collision ones. Additionally a new flags enum allows to
select which collision types to emit the events for.


bddckr:
I'm not aware of the `CollisionTracker` being used in our prefabs or scenes. Let me know and I'll change them if needed!

PR #126, opened 2018-06-18
fix(Tracking): update enum attribute to UnityFlags in collision tracker

thestonefox:
The UnityFlags attribute was recently updated but the CollisionTracker
component was still using the old UnityFlag attribute.


bddckr:
Whoops. Thanks for fixing!

PR #130, opened 2018-06-21
Chore/fixes n stuff

thestonefox:


Scripts/Process/MomentProcess.cs
```diff
@@ -17,5 +17,9 @@ public sealed class MomentProcess : MonoBehaviour
         /// </summary>
         [Tooltip("Only run the process if the process is on an active GameObject and the Component is enabled.")]
         public bool onlyProcessOnActiveAndEnabled = true;
+
+        private void OnEnable()
```
bddckr: Let's add a comment here for when we wonder about this in the future:

```c#
// This empty implementation tells Unity to draw the enabled checkbox for this component, allowing to disable the component at edit-time.
```

Scripts/Tracking/Velocity/ApplyVelocity.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Applies the velocity data from the given <see cref="VelocityTracker"/> to the given <see cref="Rigidbody"/>.
+    /// </summary>
+    public class ApplyVelocity : MonoBehaviour
+    {
+        /// <summary>
+        /// The <see cref="Rigidbody"/> to apply the velocity data to.
+        /// </summary>
+        [Tooltip("The Rigidbody to apply the velocity data to.")]
+        public Rigidbody applyTo;
```
bddckr: My suggestion is to call this `target`.

Scripts/Tracking/Velocity/ApplyVelocity.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// The <see cref="VelocityTracker "/> to receive the velocity data from.
+        /// </summary>
+        [Tooltip("The VelocityTracker to receive the velocity data from.")]
+        public VelocityTracker velocityData;
```
bddckr: This is not necessarily `data`, instead I suggest `source`. We don't need to talk about `velocity` in the member name here as it's implied by the component type we're in and the member's type.

Scripts/Tracking/Collision/CollisionList.cs
```diff
@@ -0,0 +1,89 @@
...
+            protected set;
+        } = new List<CollisionNotifier.EventData>();
+
+        /// <summary>
+        /// Determines if the collection is empty.
+        /// </summary>
+        /// <returns><see langword="true"/> if the collection is empty.</returns>
+        public virtual bool IsEmpty()
+        {
+            return (Elements.Count == 0);
```
bddckr: Do we really need to add a method for this? I'd rather have everyone type this out than introduce another method which allows to be overridden and thus break any usage of the list as the count and this empty check can get out of sync.

tl;dr: I don't see this saving us a lot so I wonder if it's worth removing it for the sake of ensuring the usage of the list is sound.

Scripts/Tracking/Collision/CollisionList.cs
```diff
@@ -0,0 +1,89 @@
...
+            }
+            EmitChanged();
+        }
+
+        /// <summary>
+        /// Emits the Changed event.
+        /// </summary>
+        public virtual void EmitChanged()
+        {
+            Changed?.Invoke(Elements);
```
bddckr: Same as above: I suggest to inline this method into its usages and remove it.

Scripts/Tracking/Collision/ListProcess/BaseListProcess.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        [Tooltip("Executes any List Action components found on the collider.")]
+        public bool executeColliderActions = false;
+
+        /// <summary>
+        /// Emitted when the <see cref="CollisionList"/> collection has been processed.
+        /// </summary>
+        public CollisionList.UnityEvent Processed = new CollisionList.UnityEvent();
+
+        protected List<CollisionNotifier.EventData> processedList = new List<CollisionNotifier.EventData>();
```
bddckr: Is this `previousList`? It seems like not all subclasses use it - should this instead be a thing in the subclasses that need it?

Scripts/Tracking/Collision/ListProcess/Action/BaseListAction.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// The base for providing an action that can be executed from a give collision.
+    /// </summary>
+    public abstract class BaseListAction : MonoBehaviour
+    {
+        /// <summary>
+        /// Executes an action on the given collision collection with the given state.
+        /// </summary>
+        /// <param name="collisionList">The collision collection at the time the action was executed.</param>
+        /// <param name="state">The state for the action.</param>
+        public abstract void Execute(List<CollisionNotifier.EventData> collisionList, bool state);
```
bddckr: This is hardcoded to take a `bool`, what if I have a List Action that wants something else?

PR #134, opened 2018-06-26
Feat/list processing

thestonefox:


Scripts/Tracking/Collision/List/SliceCollisions.cs
```diff
@@ -0,0 +1,102 @@
...
+    using UnityEngine;
+    using System.Linq;
+
+    /// <summary>
+    /// Slices a selection of the collection from the given <see cref="startIndex"/> for the given <see cref="length"/> and provides the sliced collection and the remaining collection separately.
+    /// </summary>
+    public class SliceCollisions : MonoBehaviour
+    {
+        /// <summary>
+        /// The zero-based index which to start the slice at.
```
bddckr: Remove "which". Also update the tooltip.

Scripts/Tracking/Collision/List/SliceCollisions.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        [Tooltip("The zero-based index which to start the slice at.")]
+        public uint startIndex = 0;
+        /// <summary>
+        /// The number of elements in the to slice.
```
bddckr: >The number of elements in the ~to~ slice.

Perhaps? Also update the tooltip.

Scripts/Tracking/Collision/List/SliceCollisions.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            remainingList.activeCollisions.AddRange(originalList.activeCollisions.Except(slicedList.activeCollisions));
+
+            Remained?.Invoke(remainingList);
+
+            return slicedList;
+        }
+
+        /// <summary>
+        /// Slices the collision collection and returns a new collection containing the sliced collection and provides the remaining collection as an out parameter.
```
bddckr: Let's change this to
>Slices the collision collection

as we don't need to explain what it's returning - that is handled by the other XML tags.

Scripts/Tracking/Collision/List/SortCollisionsByNearest.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            sortedList.Set(originalList);
+
+            sortedList.activeCollisions.Sort(
+            (data1, data2) =>
+                {
+                    float distance1 = Vector3.Distance(data1.collider.GetContainingTransform().position, source.position);
+                    float distance2 = Vector3.Distance(data2.collider.GetContainingTransform().position, source.position);
+                    int result = distance1.CompareTo(distance2);
```
bddckr: This looks strange as we don't introduce variables we are about to return as is if the expression is small enough. How about `return distance1.CompareTo(distance2);`?

Tests/Editor/Tracking/Collision/List/ActiveCollisionsTest.cs
```diff
@@ -0,0 +1,176 @@
...
+            Assert.IsFalse(countChangedMock.Received);
+            Assert.IsTrue(contentsChangedMock.Received);
+            Assert.IsFalse(allStoppedMock.Received);
+        }
+
+        protected CollisionNotifier.EventData GenerateCollision(out GameObject container)
+        {
+            container = new GameObject();
+            BoxCollider boxCollider = container.AddComponent<BoxCollider>();
+            Collider collider = container.GetComponent<Collider>();
```
bddckr: This is just getting the BoxCollider that was just added, thus I suggest to do
```c#
return new CollisionNotifier.EventData().Set(null, true, null, boxCollider);
```

Tests/Editor/Tracking/Collision/List/ReverseCollisionsOrderTest.cs
```diff
@@ -0,0 +1,86 @@
...
+                returnString += data.collider.name + ",";
+            }
+            return returnString;
+        }
+
+        protected CollisionNotifier.EventData GenerateCollision(out GameObject container)
+        {
+            container = new GameObject();
+            BoxCollider boxCollider = container.AddComponent<BoxCollider>();
+            Collider collider = container.GetComponent<Collider>();
```
bddckr: See my comment above.

Additionally: This is duplicated code. Perhaps this method should live in a shared class in the `Test` namespace instead?

Tests/Editor/Tracking/Collision/List/SortCollisionsByNearestTest.cs
```diff
@@ -0,0 +1,89 @@
...
+            }
+            return returnString;
+        }
+
+        protected CollisionNotifier.EventData GenerateCollision(out GameObject container, Vector3 position)
+        {
+            container = new GameObject();
+            BoxCollider boxCollider = container.AddComponent<BoxCollider>();
+            Collider collider = container.GetComponent<Collider>();
+            collider.transform.position = position;
```
bddckr: Despite this being different to the two other copies of this method we could still merge them into a method used in all these three places.

PR #135, opened 2018-06-27
Feat/pointer prefab helper

thestonefox:


Prefabs/Pointers/SharedResources/Scripts/PointerInternalSetup.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            if (objectFollow != null)
+            {
+                if (objectFollow.targetComponents.Count == 0)
+                {
+                    objectFollow.targetComponents.Add(userSetup.followTarget);
+                }
+                else
+                {
+                    objectFollow.targetComponents[0] = userSetup.followTarget;
```
bddckr: This is **replacing** the first existing `targetComponent` - is that intended?

thestonefox: Yes, it's intended. But I think a better solution is to just clear the list and do an add. There can only be one follow target

PR #139, opened 2018-07-12
feat(Rule): make rules more flexible and powerful - resolves #107

### LINKED ISSUE #107: Revamp ExclusionRule
- [x] Change the type to be a **container of** rules
- [x] Decide (and implement) on having an interface for rules like `IRule`?!
- [x] Decide on whether AND/OR'ing the rules is useful or whether we just always do one of the operations
- [x] ~Add a base rule type~
- [x] ~Add `OperationType` to the base rule type~
- [x] Add rule implementations for the different `CheckTypes`
- [x] ~Think about adding a string-component-lookup rule implementation similar to https://github.com/thestonefox/VRTK/pull/1813~ _Decided against it as a user can easily add these kinds of rules themselves now and we didn't have a lot of requests for it in v3._
- [x] Use proper types and Editor UI for the various rule implementations (e.g. use a layer picker)

bddckr:
The existing `ExclusionRule` class was used to allow filtering on
any component that needs it. This change doesn't change the intended
use case, but allows rules to be more powerful.

A new interface `IRule` defines what a rule does: It determines
whether a given object is accepted by a rule. The change from a
negated question "is this excluded" to "is this accepted" makes
reasoning about the rules and their usage easier. To use a rule in
the inspector a `RuleContainer` type is added in addition to a
helper extension method on that type, which allows calling the
`Accepts` method without checking for a `null` rule first. The
existing `ExclusionRule` is split into multiple standalone rules and
some additional rules allow to combine rules in any way possible.
Existing usages in components and tests have been updated and all
new components and types come with unit tests ensuring their
functionality is as intended.

To allow picking a type for the `AnyComponentTypeRule` a new
attribute and property drawer have been added. `RuleContainer`s can
be set and managed right in the inspector via a plus/minus button
that takes care of adding/removing the necessary rule components to
the game object while still allowing the default drag and drop
behavior.


bddckr:
Actually still need to go through all Todos in #107.

bddckr:
Todos are done and `AnyComponentRule` has been simplified by dropping the `SearchOptions` for now. There's a new issue #141 to come back to it later.

thestonefox:
The prefabs that use the previous ExclusionRule need updating to use the new rule system otherwise they have a broken script on the prefab.

 * Interactions/Interactables
    * Interactable.Touch.Grab
       * InteractionLogic/Touching/StartTouching
       * InteractionLogic/Touching/StopTouching
       * InteractionLogic/Grabbing/StartGrabbing
       * InteractionLogic/Grabbing/StopGrabbing
    * Interactable.Touch.Grab.ControlDirection + UniformScale
       * InteractionLogic/Touching/StartTouching
       * InteractionLogic/Touching/StopTouching
       * InteractionLogic/Grabbing/AnyGrabbedPressed
       * InteractionLogic/Grabbing/AnyGrabReleased

bddckr:
Ah damn, totally forgot! Will get to this ASAP tomorrow!

bddckr:
@thestonefox Done!

thestonefox:
There are still issues with the PR. The interactions no longer work when this PR is applied.

I _think_ it is due to the way the `.Accepts` method works specifically in the `CamEmit` method on the `CollisionNotifier`

it seems if there is an empty forward source validity then the `Accepts` always returns false, it would seem the `container == null` check never passes because the container isn't null. I wonder if simply having an extension method is enough to make it no longer null? Or the way it has the interface wrapping is making it not null?

Anyway, in short it doesn't work and I think it's at that fundamental level that's causing the break.

Probably worth setting up a test scene to prove it works as there is probably a false positive in the unit test.

bddckr:
Should be fixed now! Issue was a missing test against the rule container's *interface* in the extension method. I added another unit test to fail because of that in the future.

Still need to get a test scene for interactions set up on my end to be able to do that test.

thestonefox:
all seems to be working now

PR #142, opened 2018-07-16
feat(Association): extract common functionality for reusability

bddckr:
The existing `XrDeviceRelationActivator` introduced a concept that
is useful in more scenarios: Some data defines whether specified
game objects should be turned on or off. This change moves the
activation logic into a new type and introduces a new base type
describing any association. The existing association for the loaded
XR device is changed to no longer just take a string, instead it
allows using a regular expression to match the loaded XR device by
name.


thestonefox:
I think the commit message needs to be more than `associations` as its just repeating the scope.

maybe

`feat(Association): ability to determine connected VR device`

PR #143, opened 2018-07-17
 feat(Utility): disable or enable components by type - resolves #28

### LINKED ISSUE #28: Ability to enable/disable the renderers on an object
Show and hide an object (e.g. controllers) in a scene.

bddckr:
A new type has been added helping with enabling or disabling all
components on a game object that match a particular type.

### DIFF
--- a/Scripts/Utility/BehaviourEnabledSetter.cs
+++ b/Scripts/Utility/BehaviourEnabledSetter.cs
@@ -1,0 +1,48 @@
+﻿namespace VRTK.Core.Utility
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using UnityEngine;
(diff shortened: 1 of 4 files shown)


bddckr:
Requires #139.

thestonefox:
this probably needs a rebase from master after the other PRs have been merged as it's showing a commit from a different PR

thestonefox:
Also, is it worth having this throw an event when something is disabled and something is enabled?

E.g. i hide my controller, i may want to let things know that it is hidden (and visible again)

Could just be an empty UnityEvent can't think of any payload data that it should care about

bddckr:
Decided against events until we have a better idea of the use case they're needed for.

PR #155, opened 2018-08-14
Fix/buggles

thestonefox:


Scripts/Tracking/SurfaceLocator.cs
```diff
@@ -146,19 +147,54 @@ protected virtual bool CastRay(Vector3 givenOrigin, Vector3 givenDirection)
...
+                {
+                    SetSurfaceData(collision);
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Sets the SurfaceData collision information.
```
bddckr: `<see...` `SurfaceData` instead?

PR #161, opened 2018-08-28
Feat/intertactions - fixes #113 fixes #148

### LINKED ISSUE #113: [Header] position breaks XML docs
`[Header]` attributes are giving us issues with the XML docs. This:
```c#
        [Header("Object Follow Settings")]

        /// <summary>
        /// The <see cref="Transform"/> properties to follow.
        /// </summary>
        [UnityFlags]
        [Tooltip("The Transform properties to follow.")]
        public TransformProperties follow = TransformProperties.Position | TransformProperties.Rotation | TransformProperties.Scale;
```

needs to be this instead to make the docs work:
```c#
        /// <summary>
        /// The <see cref="Transform"/> properties to follow.
        /// </summary>
        [Header("Object Follow Settings")]

        [UnityFlags]
        [Tooltip("The Transform properties to follow.")]
        public TransformProperties follow = TransformProperties.Position | TransformProperties.Rotation | TransformProperties.Scale;
```

Basically it needs to be
```
XML docs
attributes
member (field or whatever)
```

I don't like the look of the proposed solution at all as it looks like we're documenting the header now... Removing and newline kinda is better but that looks like the header is part of the object (which it technically is as attributes annotate members...).

One can merge (some or all) attributes, but that also looks wrong and makes it harder to see them:
```c#
        /// <summary>
        /// The <see cref="Transform"/> properties to follow.
        /// </summary>
        [Header("Object Follow Settings")]
        [UnityFlags, Tooltip("The Transform properties to follow.")]
        public TransformProperties follow = TransformProperties.Position | TransformProperties.Rotation | TransformProperties.Scale;
```

Need to add whatever we decide on to the CONTRIBUTING file, too.

### LINKED ISSUE #148: Interaction grabbing is sometimes possible when it shouldn't
Raise another PR with a fix to the interactions grab issue - just needs glue code in the top grab script to pass down the list of valid interactors to a `ListContainsRule`.

thestonefox:


thestonefox:
This is probably going to be a bit of hit and miss with what's worth reviewing.

Basically it started off as me going down one route to get things working and I built some scripts that are probably useful for the future (like the GameObjectComparer) but then later on they don't get used at all, but I was in such a kerfuffle I just left the commit in just so nothing broke and made me cry.

Some commits are distinct and not really to do with this feature but hey-ho i bundled them in anyway for lols!

Scripts/Tracking/GameObjectComparer.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// <param name="givenGameObject">The <see cref="GameObject"/> to compare against the <see cref="compareTo"/>.</param>
+        /// <returns><see langword="true"/> if the comparison is successful.</returns>
+        public virtual bool Compare(GameObject givenGameObject)
+        {
+            if (!IsValid())
+            {
+                return false;
+            }
+
+            bool result = (compareTo == givenGameObject);
```
bddckr: Instead of just having a `GameObject compareTo` I propose using an `RuleContainer rule` and implementing an `IRule` (inheriting from `BaseGameObjectRule`) that returns whether the passed GameObject is equal to a set one.

This won't change the intended usage here but results in a reusable component. I propose the type name changes to `RuleTester` or something.

Scripts/Data/Collection/GameObjectSet.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// Clears all elements from the set.
+        /// </summary>
+        public virtual void ClearElements()
+        {
+            if (!isActiveAndEnabled)
+            {
+                return;
+            }
+
+            Elements.Clear();
```
bddckr: Missing
```c#
BecameEmpty?.Invoke(element);
```

Scripts/Data/Collection/GameObjectSet.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            if (!isActiveAndEnabled)
+            {
+                return false;
+            }
+
+            bool result = Elements.Contains(element);
+            if (result)
+            {
+                ElementFound?.Invoke(element);
```
bddckr: These events are strange as there is no connection between the caller of the method and the receiver of the event.

I assume this is needed for wiring things up in the Unity inspector and we can't instead switch those to be done in a script just utilizing the return value of the method?

Scripts/Data/Collection/GameObjectSet.cs
```diff
@@ -0,0 +1,129 @@
...
+                ElementNotFound?.Invoke(element);
+            }
+            return result;
+        }
+
+        /// <summary>
+        /// Determines if the given element is contained within the set and emits the appropriate event based on the result.
+        /// </summary>
+        /// <param name="element">The element to search for.</param>
+        public virtual void DoContains(GameObject element)
```
bddckr: Same as above.

Scripts/Extension/GameObjectExtensions.cs
```diff
@@ -29,5 +29,34 @@ public static void TrySetActive(this GameObject gameObject, bool state)
...
         }
+
+        /// <summary>
+        /// Attempts to retreive the <see cref="Rigidbody"/> or if one is not found then optionally search children then search parents.
+        /// </summary>
+        /// <param name="gameObject">The reference <see cref="GameObject"/> to search on.</param>
+        /// <param name="searchAncestors">Optionally searches all ancestors in the hierarchy for a rigidbody.</param>
+        /// <param name="searchDescendants">Optionally searches all descendants in the hierarchy for a rigidbody.</param>
+        /// <returns>The found <see cref="Rigidbody"/></returns>
+        public static Rigidbody FindRigidbody(this GameObject gameObject, bool searchDescendants = true, bool searchAncestors = false)
```
bddckr: Change this to be generic instead:
```c#
public static T FindComponent<T>(this GameObject gameObject, bool searchDescendants = true, bool searchAncestors = false) where T : Component
```
Update the documentation, tests, type names and usages accordingly.

**This should probably be replacing the `TryGetComponent` method then as it's a more powerful replacement. Beware of the default parameter values, though, as the `TryGetComponent` method didn't act like that so far.**

Scripts/Extension/GameObjectExtensions.cs
```diff
@@ -8,13 +8,14 @@
...
-        /// Attempts to retrieve the <see cref="Component"/> from a given <see cref="GameObject"/>.
+        /// Attempts to retrieve the component from a given <see cref="GameObject"/>.
         /// </summary>
-        /// <param name="gameObject">The <see cref="GameObject"/> to retrieve the <see cref="Component"/> from.</param>
-        /// <returns>The <see cref="Component"/> if one exists on the given <see cref="GameObject"/>.</returns>
-        public static Component TryGetComponent(this GameObject gameObject)
+        /// <typeparam name="T">The component type to retrieve.</typeparam>
+        /// <param name="gameObject">The <see cref="GameObject"/> to retrieve from.</param>
+        /// <returns>The component if one exists.</returns>
+        public static T TryGetComponent<T>(this GameObject gameObject)
```
bddckr: Missing type constraint:
```c#
public static T TryGetComponent<T>(this GameObject gameObject) where T : Component
```

.github/CONTRIBUTING.md
```diff
@@ -188,22 +207,52 @@ lines included in `<remarks>`.
...
-  * Tooltip attribute
+  * Parameter attributes
 
 ```
 /// <summary>
 /// Description here.
 /// </summary>
-[Optional-Attributes...]
-[Tooltip("Description here.")]
+[Tooltip("Description here."), InternalSetting, SerializeField]
```
bddckr: There is no need for `SerializeField` on `public` fields and therefore it should not be listed.

Either change the example here to `protected` or drop the `SerializeField` annotation.

.github/CONTRIBUTING.md
```diff
@@ -188,22 +207,52 @@ lines included in `<remarks>`.
...
+the `#region` directive to denote the block assoicated.
+The following shows a valid and invalid example:
+
+  > **Incorrect:**
+  ```
+  [Header("My Header")]
+  /// <summary>
+  /// Description here.
+  /// </summary>
+  [Tooltip("Description here."), InternalSetting, SerializeField]
```
bddckr: Same as above.

.github/CONTRIBUTING.md
```diff
@@ -188,22 +207,52 @@ lines included in `<remarks>`.
...
+  public Type myType;
+  ```
+
+  > **Correct:**
+  ```
+  #region My Header
+  /// <summary>
+  /// Description here.
+  /// </summary>
+  [Header("My Header"), Tooltip("Description here."), InternalSetting, SerializeField]
```
bddckr: Same as above.

Scripts/Tracking/GameObjectComparer.cs
thestonefox: I'm wondering whether to just remove the GameObjectComparer commit completely, it's not used for anything right now

Scripts/Data/Collection/GameObjectSet.cs
thestonefox: I pondered this one.

You can't call `BecameEmpty` because you have no element to invoke with.

The BecameEmpty invokes with the last element that was removed from the set.

Now if you do `Clear` you have no last element (as it's a set so it's not ordered)

Whilst I agree you'd expect BecameEmpty to fire when clear happens, what should be the emitted element? null maybe?

Scripts/Data/Collection/GameObjectSet.cs
thestonefox: Not sure what you mean here.

The event just rethrows the thing that you gave it so you can use it for chaining.

E.g.

GameObject -> is it in the set? -> if yes then do x with GameObject if no then do y with GameObject

Scripts/Tracking/GameObjectComparer.cs
bddckr: True... Not too much work to recreate once we need it I think!

Scripts/Extension/GameObjectExtensions.cs
thestonefox: Isn't there a thing where `GetComponent` can also get things that don't derive from component? Like Renderers and stuff like that?

.github/CONTRIBUTING.md
thestonefox: I'll change the example to protected

Scripts/Data/Collection/GameObjectSet.cs
bddckr: Yeah `null` is better than not firing it I'd say. Maybe just document that event data to be the object that was removed and `null` in the case of emptying the whole collection?

If we don't need the element in the event we could remove it (but I assume you had it in for a reason 😛).

Scripts/Data/Collection/GameObjectSet.cs
bddckr: 1. Three things listen for the event.
2. _Someone_ asks for whether a given object is contained.
3. All listeners get an answer despite not even being the one that asked.

Scripts/Extension/GameObjectExtensions.cs
bddckr: They all inherit from `Component`. Our issue was that a `Renderer` for example doesn't inherit from _`Behaviour`_.

Scripts/Data/Collection/GameObjectSet.cs
thestonefox: You wire up your events in the inspector to be things that do something when that event is fired.

So the usage is

Check set contains ObjectA
If it does contain it then do x,y,z (the three listeners)

The listener doesn't do the asking, the listener is just set up to respond to the question

Scripts/Data/Collection/GameObjectSet.cs
thestonefox: Yeah you need to know the element that was last emptied out of the set for example when an interactor is finally removed from the set you want to know what GameObject was removed so you can call an event like `Ungrabbed`

PR #164, opened 2018-09-03
feat(Follow): add rigidbody rotation by force follow modifier property

thestonefox:
A new follow modifier property dealing with rotation has been added
that applies a force at a given position to a rigidbody which creates
a torque providing a rotation to the object.

This is useful for controlling objects that are attached on a hinge
joint.

As this is using actual forces it cannot be tested via the editor
tests.

### DIFF
--- a/Scripts/Tracking/Follow/Modifier/Property/Rotation/RigidbodyForceAtPosition.cs
+++ b/Scripts/Tracking/Follow/Modifier/Property/Rotation/RigidbodyForceAtPosition.cs
@@ -1,0 +1,63 @@
+﻿namespace VRTK.Core.Tracking.Follow.Modifier.Property.Rotation
+{
+    using UnityEngine;
+    using VRTK.Core.Extension;
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


Scripts/Tracking/Follow/Modifier/Property/Rotation/RigidbodyForceAtPosition.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            cachedTargetRigidbody = (cachedTargetRigidbody == null || target != cachedTarget ? target.TryGetComponent<Rigidbody>(true) : cachedTargetRigidbody);
+            cachedTarget = target;
+
+            if (cachedTargetRigidbody == null || source == null || attachmentPoint == null)
+            {
+                return;
+            }
+
+            Vector3 rotateForce = source.transform.position - attachmentPoint.transform.position;
```
bddckr: Reads like a bool - what about `rotationForce` instead?

PR #165, opened 2018-09-03
feat(Modification): option to lock look at z axis rotation

thestonefox:
The DirectionModifier can now have the look at target prevented from
applying z axis rotation to the grabbed object. This can provide a
more natural direction control in certain circumstances.

This solution is taken from VRTK v3 and was based largely on the
solution provided by @DrHibbitts


thestonefox:
in this regard you only ever want to fix the z rotation if you locked y/x rotation then it would never look at the target :)

bddckr:
Yeah since it's done in world-space that makes sense I guess. Good to merge!

PR #167, opened 2018-09-06
Feat/behaviour state setter - fixes #28

### LINKED ISSUE #28: Ability to enable/disable the renderers on an object

thestonefox:


Scripts/Event/RestrictableSingleEventProxyEmitter.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// Determines whether the received payload is valid to be re-emitted.
+        /// </summary>
+        [Tooltip("Determines whether the received payload is valid to be re-emitted.")]
+        public RuleContainer receiveValidity;
+
+        /// <summary>
+        /// Gets the target for the validity check.
+        /// </summary>
+        /// <returns>The target to check on.</returns>
+        protected abstract object GetAcceptsTarget();
```
bddckr: `GetPotentialTarget` or `GetTargetToCheck` instead perhaps? Kinda ready wrong with the `Accepts` in there as it talks about the API's name it's used for rather than describing the returned object.

PR #168, opened 2018-09-07
feat(Rule): match against a collection of rules and emit event

thestonefox:
The RulesMatcher allows a for a List of rules to be provided with
an associated event for each rule. When the `Match` method is called
with a source GameObject, any rule in the list that matches based
on the given GameObject will emit the associated event.

### DIFF
--- a/Scripts/Rule/RulesMatcher.cs
+++ b/Scripts/Rule/RulesMatcher.cs
@@ -1,0 +1,56 @@
+﻿namespace VRTK.Core.Rule
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using System;
+    using System.Linq;
+    using System.Collections.Generic;
+    using VRTK.Core.Extension;
+
+    /// <summary>
+    /// Matches a given object against a collections of rules and emits the associated event.
+    /// </summary>
+    public class RulesMatcher : MonoBehaviour
+    {
+        /// <summary>
+        /// The rule and event association that can be matched against.
+        /// </summary>
+        [Serializable]
+        public class Element
+        {
+            /// <summary>
+            /// The rule to match against.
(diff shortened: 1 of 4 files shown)


Scripts/Rule/RulesMatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// </summary>
+        /// <param name="source">The source to provide to the rule for validity checking.</param>
+        public virtual void Match(GameObject source)
+        {
+            if (!isActiveAndEnabled)
+            {
+                return;
+            }
+
+            foreach (Element element in elements.Where(target => target.rule.Accepts(source)))
```
bddckr: `elements.Where(...` ➡️ `elements.EmptyIfNull().Where(...`

Scripts/Rule/RulesMatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// A collection of rules to potentially match against.
+        /// </summary>
+        [Tooltip("A collection of rules to potentially match against.")]
+        public List<Element> elements = new List<Element>();
+
+        /// <summary>
+        /// Attempts to match the given <see cref="GameObject"/> to the rules within the <see cref="elements"/> collection. If a match occurs then the appropriate event is emitted.
+        /// </summary>
+        /// <param name="source">The source to provide to the rule for validity checking.</param>
+        public virtual void Match(GameObject source)
```
bddckr: I wonder if this should take `object` (the .NET one, not the Unity one) instead. Rules allow anything currently so it may be worth checking if Unity allows passing let's say `GameObject` when wiring up to this in the inspector.

thestonefox:
updated with the requested changes.

PR #169, opened 2018-09-11
Feat/grabbing with pointer - fixes #41

### LINKED ISSUE #41: Allow pointers to interact with objects (e.g. grab objects with a pointer)
This should be totally decoupled and not baked in like it currently is.

thestonefox:


Scripts/Cast/FixedLineCast.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// The current length of the cast.
+        /// </summary>
+        [Tooltip("The current length of the cast.")]
+        public float currentLength = 1f;
+
+        /// <summary>
+        /// Sets the length of the cast to the given length.
+        /// </summary>
+        /// <param name="length">The new length of the cast.</param>
+        public virtual void SetLength(float currentLength)
```
bddckr: Should this be called `SetCurrentLength` instead?

I see why the field isn't just `length` (potential confusion with the superclass' length field), but these setter methods should probably be using the same naming compared to the field.

Scripts/Cast/FixedLineCast.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            this.currentLength = currentLength;
+        }
+
+        /// <summary>
+        /// Sets the length of the cast from the given event data.
+        /// </summary>
+        /// <param name="data">The data to set the length from.</param>
+        public virtual void SetLength(EventData data)
```
bddckr: Same as above: Should this be called `SetCurrentLength` instead?

PR #172, opened 2018-09-11
feat(Interaction): add grab/ungrab helper methods to interactors

thestonefox:
The InteractorFacade now has a Grab and Ungrab helper method that
allows an Interactable to be passed to a specific Interactor to
initiate a grab (or alternatively ungrab).

This makes it possible to initiate a grab/ungrab from either the
Interactor or the Interactable.


thestonefox:
Removed the whole invoke thing now. Provides way of always going through the Interactor to do the grab/ungrab

toggle grab type on an interactable is handled by the interactor code.

It's possible to pass custom collision data to the Interactor grab if a precision grab is necessary but it has to be manually provided (if the objects aren't currently touching). If they are touching then the precision grab should work as it has a collision point already.

Only supports single grabbing/ungrabbing of objects.

PR #176, opened 2018-09-16
fix(CameraRig): don't change time settings at edit-time

bddckr:
The configuration component for the Unity XR settings was changing
the project time settings while in the editor and not playing. Since
the XR device is not loaded at edit-time this resulted in wrong time
settings. The fix is to only run the setting logic when playing.


thestonefox:
Maybe everywhere we use `OnValidate` we should also make sure it's never called at edit time too? If you think that's a good idea lemme know and I'll raise a ticket

bddckr:
@thestonefox Depends I think - there could be a case of a component wanting to change something right there in the Editor. I don't think we have that need right now and we should definitely look through other use cases of `OnValidate`!

PR #199, opened 2018-11-10
feat(Velocity): calculate with optional relative to object - fixes #197

### LINKED ISSUE #197: Add relative to GameObject to velocity estimator
The existing velocity estimator works in world space therefore if the controller is moved because the play area moves then it will report a controller velocity even if the controller hasn't moved irl.

The position of the tracked object needs to take into consideration a relative object, so in the case of the controller it would be relative to the play area

thestonefox:
The VelocityEstimator now allows for a relativeTo GameObject that
can be used to determine a relative velocity rather than a global
velocity.

This is useful if the tracked object is also being moved due to
a parent object being moved.

The CameraRig prefabs have been updated so the controller velocity
is always tracked relative to the parent play area.

### DIFF
--- a/Scripts/Tracking/Velocity/VelocityEstimator.cs
+++ b/Scripts/Tracking/Velocity/VelocityEstimator.cs
@@ -2,6 +2,7 @@
 {
     using UnityEngine;
     using System;
+    using VRTK.Core.Extension;
 
     /// <summary>
     /// Takes average samples of a <see cref="Transform.position"/> and <see cref="Transform.rotation"/> and uses this cached data to estimate velocity and angular velocity.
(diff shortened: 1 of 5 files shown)


Scripts/Tracking/Velocity/VelocityEstimator.cs
```diff
@@ -191,12 +218,15 @@ protected virtual void ProcessEstimation()
         /// <param name="factor">The multiplier to apply to the transform difference.</param>
         protected virtual void EstimateVelocity(float factor)
         {
-            Vector3 currentPosition = (source != null ? source.transform.position : Vector3.zero);
+            Vector3 currentRelativePosition = relativeTo.TryGetPosition();
+            Vector3 relativeDeltaPosition = currentRelativePosition - previousRelativePosition;
+            Vector3 currentPosition = source.TryGetPosition();
             if (velocitySamples.Length > 0)
```
bddckr: Negate to `== 0` and create an early return with that. This way we don't do the math above this if without using the results.

Scripts/Tracking/Velocity/VelocityEstimator.cs
```diff
@@ -225,6 +257,7 @@ protected virtual void EstimateAngularVelocity(float factor)
                 int sampleIndex = currentSampleCount % angularVelocitySamples.Length;
```
bddckr: (I actually wanted to comment on the `if` line.)

Same as above: Negate to `== 0` and make it an early return right at the start of the method, otherwise we're doing all this math without using it.

PR #205, opened 2018-11-11
Feat/process utilization - resolves #204

### LINKED ISSUE #204: Allow reducing how often an IProcessable runs

bddckr:


Scripts/Process/Moment/MomentProcess.cs
```diff
@@ -17,10 +17,71 @@ public sealed class MomentProcess : MonoBehaviour
...
+        /// Calls <see cref="IProcessable.Process"/> on <see cref="process"/> if <see cref="Utilization"/> allows.
+        /// </summary>
+        public virtual void Process()
+        {
+            if (Utilization < float.Epsilon)
+            {
+                return;
+            }
+
+            if (process != null && (!onlyProcessOnActiveAndEnabled || isActiveAndEnabled) && counter == delay)
```
thestonefox: The exception hasn't been carried over in case the interface process is null. Should we be doing that here too?

bddckr: I removed it as accessing it will give the same null ref exception and Unity logs the member name these days.

Scripts/Process/Moment/MomentProcess.cs
```diff
@@ -17,10 +17,71 @@ public sealed class MomentProcess : MonoBehaviour
...
+        /// </summary>
+        public float Utilization
+        {
+            get
+            {
+                return utilization;
+            }
+            set
+            {
+                utilization = Mathf.Clamp(value, 0f, 1f);
```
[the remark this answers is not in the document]
bddckr: Done and force-pushed.

PR #208, opened 2018-11-12
feat(Locomotion): allow teleport offset to be optional - resolves #207

### LINKED ISSUE #207: Allow Teleporter prefabs to not teleport with an offset between the playarea and headset.
At the moment, the teleporter prefabs will always take into consideration the offset of the headset within the play area, so any new teleported position will be always based on where the headset is located and the play area is moved with that offset considered.

However, it may be required when teleporting to move the target to the exact point of teleport without any offset.

Also, it's worth changing the facade parameters to make them a bit more generic.

I propose:

PlayAreaAlias -> Target
HeadsetAlias -> Offset

Also now need an optional option to say when to use that offset. the offset of Headset **needs** to be used in the SnapToFloor PropertyApplier so it's always putting you on the floor where you're standing, but in the HeightAdjustTeleporter PropertyApplier you only want the offset if you want the current behaviour, so a simple bool of:

ApplyOffsetOnTeleport (or something) which is default true.

No point in providing two different aliases as this would just break things as the aliases need to be the same, or at least SnapToFloor populated and HeightAdjust empty.

thestonefox:
The Teleporter would previously always apply an offset of the
HeadsetAlias to the PlayAreaAlias when moving it but there can be
instances where the PlayAreaAlias centre is wanted to be moved to
the teleport point.

This is now possible by optionally setting the Offset parameter
on the TransformPropertyApplier inside the HeightAdjustTeleporter
GameObject.

The parameter fields have also changed to be more generic.

PlayAreaAlias becomes Target
HeadsetAlias becomes Offset

And a new bool, of `OnlyOffsetFloorSnap` allows for the offset
on teleport to be optional. If the setting is true then the Offset
is only applied when doing floor snapping, which has to always be
applied and cannot be optional.

### DIFF
--- a/Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterFacade.cs
+++ b/Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterFacade.cs
@@ -13,15 +13,20 @@
     {
         #region Teleporter Settings
         /// <summary>
-        /// The alias for the CameraRig Play Area.
+        /// The target to move to the teleported position.
         /// </summary>
-        [Header("Teleporter Settings"), Tooltip("The alias for the CameraRig Play Area.")]
-        public GameObject playAreaAlias;
+        [Header("Teleporter Settings"), Tooltip("The target to move to the teleported position.")]
+        public GameObject target;
         /// <summary>
-        /// The alias for the CameraRig Headset.
+        /// The offset to compensate the teleported target position by for both floor snapping and position movement.
         /// </summary>
-        [Tooltip("The alias for the CameraRig Headset.")]
-        public GameObject headsetAlias;
+        [Tooltip("The offset to compensate the teleported target position by for both floor snapping and position movement.")]
+        public GameObject offset;
+        /// <summary>
+        /// Determines if only the floor snap should only be compensated by the <see cref="offset"/> or whether the teleported target position should also be compensated by the <see cref="offset"/>.
+        /// </summary>
+        [Tooltip("Determines if only the floor snap should only be compensated by the offset or whether the teleported target position should also be compensated by the offset.")]
+        public bool onlyOffsetFloorSnap;
         /// <summary>
         /// The <see cref="CameraList"/> of scene <see cref="Camera"/>s to apply a fade to.
         /// </summary>
--- a/Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterInternalSetup.cs
+++ b/Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterInternalSetup.cs
@@ -7,7 +7,6 @@
     using VRTK.Core.Visual;
     using VRTK.Core.Data.Type;
     using VRTK.Core.Extension;
-    using VRTK.Core.Rule;
 
     /// <summary>
     /// Sets up the Teleport Prefab based on the provided user settings.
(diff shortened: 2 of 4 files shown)


Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterFacade.cs
```diff
@@ -13,15 +13,20 @@ public class TeleporterFacade : MonoBehaviour
     {
         #region Teleporter Settings
         /// <summary>
-        /// The alias for the CameraRig Play Area.
+        /// The target to move to the teleported position.
         /// </summary>
-        [Header("Teleporter Settings"), Tooltip("The alias for the CameraRig Play Area.")]
-        public GameObject playAreaAlias;
+        [Header("Teleporter Settings"), Tooltip("The target to move to the teleported position."), RequiredSetting]
```
bddckr:
```suggestion
        [Header("Teleporter Settings"), Tooltip("The target to move to the teleported position.")]
```

Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterFacade.cs
```diff
@@ -13,15 +13,20 @@ public class TeleporterFacade : MonoBehaviour
...
         /// <summary>
-        /// The alias for the CameraRig Headset.
+        /// The offset to compensate the teleported target position by for both floor snapping and position movement.
         /// </summary>
-        [Tooltip("The alias for the CameraRig Headset.")]
-        public GameObject headsetAlias;
+        [Tooltip("The offset to compensate the teleported target position by for both floor snapping and position movement."), RequiredSetting]
+        public GameObject offset;
+        /// <summary>
+        /// Determines if the only the floor snap should only be compensated by the <see cref="offset"/> or whether the teleported target position should also be compensated by the <see cref="offset"/>.
```
bddckr:
```suggestion
        /// Determines if the floor snap should only be compensated by the <see cref="offset"/> or whether the teleported target position should also be compensated by the <see cref="offset"/>.
```

Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterFacade.cs
```diff
@@ -13,15 +13,20 @@ public class TeleporterFacade : MonoBehaviour
...
+        /// </summary>
+        [Tooltip("Determines if the only the floor snap should only be compensated by the offset or whether the teleported target position should also be compensated by the offset.")]
```
bddckr:
```suggestion
        [Tooltip("Determines if the floor snap should only be compensated by the offset or whether the teleported target position should also be compensated by the offset.")]
```

Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterFacade.cs
```diff
@@ -13,15 +13,20 @@ public class TeleporterFacade : MonoBehaviour
...
+        public bool onlyOffsetFloorSnap = false;
```
bddckr:
```suggestion
        public bool onlyOffsetFloorSnap;
```
So far we seem to favor implicit default values if they are the same.

Prefabs/Locomotion/Teleporters/SharedResources/Scripts/TeleporterInternalSetup.cs
```diff
@@ -52,38 +51,17 @@ public class TeleporterInternalSetup : MonoBehaviour
         [Tooltip("The Transform Property Applier collection to set aliases on.")]
         public List<TransformPropertyApplier> transformPropertyApplierAliases = new List<TransformPropertyApplier>();
         /// <summary>
+        /// The <see cref="TransformPropertyApplier"/> collection to ignore offsets on.
+        /// </summary>
+        [Tooltip("The Transform Property Applier collection to to ignore offsets on.")]
```
bddckr:
```suggestion
        [Tooltip("The Transform Property Applier collection to ignore offsets on.")]
```

PR #214, opened 2018-11-15
refactor(Structure): remove use of Base from abstract classes - fixes #198

### LINKED ISSUE #198: Remove "Base" names from abstract classes
The only reason we did have different naming for those is because we wanted to ensure users understand it.

But since the inspector won't show abstract classes and the usage of those directly results in compiler errors we are good to drop the additional name suffix/prefix.

thestonefox:
The abstract classes don't need to use the word Base to determine they
are the base class. This was originally done to prevent users getting
confused in the Unity Inspector, but it doesn't show abstract classes
anyway so it makes little sense to force the term Base.


thestonefox:
Whilst updating the prefabs, a number of them now have the `utilization` property on the `MomentProcess` stored on them, so whilst they're not technically part of this commit it's at least a worthy thing to get in.

PR #215, opened 2018-11-16
refactor(Collection): rename observable collections

thestonefox:
The Stack and Set collections have now been brought into line with
the ObservableList by both now having a base class with the name
Observable included:

 * ObservableStack
 * ObservableSet

And the previous classes now become smaller concrete classes that
extend those bases to deliver a GameObjectObservableStack and a
GameObjectObservableSet.

The prefabs have been updated with the new script names.


thestonefox:
I forced pushed the change as it was just a single fix.

PR #218, opened 2018-11-17
refactor(Transformation): convert multipliers to be aggregators - fixes #217

### LINKED ISSUE #217: Convert `Data/Type/Transformation/` Multipliers to extend `CollectionAggregator`
* FloatMultiplier
* Vector2Multiplier
* Vector3Multiplier

Can all extend `CollectionAggregator` as they are all performing an aggregate function.

thestonefox:
The various Multipler transformations now extend CollectionAggregator
as multiplication is an aggregate function across a collection of
values.

Now it's possible to multiple a range of float/vector2/vector3 values
as they extend the base aggregator.

The prefabs have been updated and the general rule is the dynamic
multiplication value is always index 0 and the multipler is always
set to index 1. This way the event updating the dynamic value never
has to bother setting the current index in the collection.

### DIFF
--- a/Scripts/Data/Type/Transformation/CollectionAggregator.cs
+++ b/Scripts/Data/Type/Transformation/CollectionAggregator.cs
@@ -24,38 +24,28 @@
         /// </summary>
         public int CurrentIndex
         {
-            get;
-            protected set;
+            get { return _currentIndex; }
+            set { _currentIndex = PreprocessCurrentIndex(value); }
         }
+        private int _currentIndex;
 
         /// <summary>
-        /// Sets the collection index that will store a given input value. The given index will be clamped to the range of valid indexes.
-        /// </summary>
-        /// <param name="index">The index to set the collection value at.</param>
-        public virtual void SetIndex(int index)
-        {
-            index = (index >= 0 ? index : collection.Count + index);
-            CurrentIndex = Mathf.Clamp(index, 0, collection.Count - 1);
-        }
-
-        /// <summary>
-        /// Updates the current collection index value with the given <see cref="TInput"/>.
+        /// Updates the collection at the <see cref="CurrentIndex"/> with the given <see cref="TInput"/>.
         /// </summary>
         /// <param name="input">The element to update the collection with.</param>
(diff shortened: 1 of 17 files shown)


Scripts/Data/Type/Transformation/CollectionAggregator.cs
```diff
@@ -98,8 +88,29 @@ protected override TOutput Process(TInput input)
...
+            SetElement(input);
             return ProcessCollection();
         }
+
+        /// <summary>
+        /// Preprocesses the value to be set in <see cref="CurrentIndex"/>
+        /// </summary>
+        /// <param name="value">The value to process.</param>
+        /// <returns>The processed value</returns>
+        protected virtual int PreprocessCurrentIndex(int value)
```
bddckr: It looks like this method can be removed and the property above just uses `ProcessIndex` like everything else.

Scripts/Data/Type/Transformation/CollectionAggregator.cs
```diff
@@ -98,8 +88,29 @@ protected override TOutput Process(TInput input)
...
+        {
+            return ProcessIndex(value);
+        }
+
+        /// <summary>
+        /// Processes the given index into a valid index.
+        /// </summary>
+        /// <param name="index">The value to process.</param>
+        /// <returns>The processed index.</returns>
+        protected virtual int ProcessIndex(int index)
```
bddckr:
```suggestion
        protected virtual int WrapAroundAndClamp(int index)
```
Just a suggestion.

Scripts/Data/Type/Transformation/CollectionAggregator.cs
thestonefox: This is in anticipation of the code weaving stuff.

The `PreprocessCurrentIndex` will be the method that's auto woven into the `CurrentIndex` field.

But the actual function being done inside `PreprocessCurrentIndex` is used elsewhere so it's separated out.

thestonefox:
Force pushed the changes

PR #220, opened 2018-11-19
feat(Locomotion): ability to move a set distance in a given direction - fixes #46 - fixes #47

### LINKED ISSUE #46: Warp object a certain distance in a certain direction based on Vector2 input
The concept of player warping but more generic

### LINKED ISSUE #47: Rotate an object based on given input
This will form the rotate camera rig on touchpad/thumbstick concept,

thestonefox:
The AxesToVector3 prefab now has the ability to accept a direction on
a given axis and to move a specified distance on the first detection
of that axis direction.

This will enable the concept of warp walking, where the joystick input
is moved in a direction and on the first push it will move the player
forward a specified amount of position with a blink transition.

### DIFF
--- a/Scripts/Data/Type/Transformation/FloatToBoolean.cs
+++ b/Scripts/Data/Type/Transformation/FloatToBoolean.cs
@@ -26,7 +26,7 @@
         /// <summary>
         /// The bounds in which the float must be to be considered a positive boolean.
         /// </summary>
-        [Tooltip("The bounds in which the float must be to be considered a positive boolean."), MinMaxRange(0f, 1f), SerializeField]
+        [Tooltip("The bounds in which the float must be to be considered a positive boolean."), MinMaxRange(-1f, 1f), SerializeField]
         protected FloatRange positiveBounds = new FloatRange(0f, 1f);
 
         /// <summary>
@@ -48,7 +48,7 @@
         }
 
         /// <summary>
-        /// Transforms the given input bool to the float equavalent value.
+        /// Transforms the given input float to the equivalent bool value.
         /// </summary>
         /// <param name="input">The value to transform.</param>
         /// <returns>The transformed value.</returns>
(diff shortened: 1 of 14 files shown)


Scripts/Data/Type/Transformation/FloatToDirection.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        [Tooltip("The bounds in which the float must be outside of to be considered a valid direction."), MinMaxRange(-1f, 1f), SerializeField]
+        protected FloatRange deadzone = new FloatRange(-1f, 1f);
+        /// <summary>
+        /// The number of decimal places to round the output to.
+        /// </summary>
+        [Tooltip("The number of decimal places to round the output to.")]
+        public int roundingPrecision = 0;
+        /// <summary>
+        /// Will continually repeat the transform process with every value that is given outside of the deadzone bounds.
```
bddckr:
```suggestion
        /// Whether to continually repeat the transform process with every given value that is outside of the deadzone bounds.
```

Scripts/Data/Type/Transformation/FloatToDirection.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        [Tooltip("Will continually repeat the transform process with every value that is given outside of the deadzone bounds.")]
```
bddckr:
```suggestion
        [Tooltip("Whether to continually repeat the transform process with every given value that is outside of the deadzone bounds.")]
```

Scripts/Data/Type/Transformation/FloatToDirection.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using System;
+    using VRTK.Core.Data.Attribute;
+
+    /// <summary>
+    /// Transforms a float value to a distinct rounded direction when the float is outside of the specified deadzone.
+    /// </summary>
+    /// <example>
```
bddckr: I honestly don't get how it works just yet so I don't know if these examples are correct (or whether we should maybe list another set of them for another deadzone).

Scripts/Data/Type/Transformation/FloatToBoolean.cs
```diff
@@ -48,7 +48,7 @@ public virtual void SetPositiveBounds(Vector2 positiveBounds)
         }
 
         /// <summary>
-        /// Transforms the given input bool to the float equavalent value.
+        /// Transforms the given input float to the equavalent bool value.
```
bddckr:
```suggestion
        /// Transforms the given input float to the equivalent bool value.
```

PR #232, opened 2018-11-29
Fix/interactable touching - fixes #230

### LINKED ISSUE #230: Collision Notifier/Interactor sends touch/untouch event for every collision/trigger enter/exit for compound colliders
The Collision Notifier picks up every collision started/stopped event when moving through colliders on a compound collider GameObject

e.g. setup

MyObj (contains rigidbody)
--+ Box (box collider)
--+ Ball (sphere collider)

if the interactor touches MyObj via the box collider you get a `touch` event then when you continue to touch the MyObj but move the interactor to the other collider in the compound set you get an `untouch` event on the box collider and get a fresh `touch` event on the sphere collider. Even though you've not actually stopped and started touching the same object because you're just moving around the compound colliders.

This has a further knock on effect that an interactor will trigger the touch/untouch events on an interactable even when the object hasn't been untouched.

If you move the interactor through an interactable with multiple compound colliders you can get it to fire an untouch event even though the interactable hasn't stopped being touched.

What needs to happen is only consider a touch/untouch even if the entire containing object (i.e. the GameObject that holds the rigidbody) is untouched.

Not sure how to handle this.

There needs to be some way of knowing that when you start touching an object on `colliderA` if you stop touching `colliderA` but start touching `colliderB` on the object then you haven't untouched the object.

V3 handled this in a way of checking if the new touching collider was part of the existing touching object(s) and if it was it didn't bother doing the touch again:
https://github.com/thestonefox/VRTK/blob/master/Assets/VRTK/Source/Scripts/Interactions/Interactors/VRTK_InteractTouch.cs#L293

It would maintain a list of the active touched colliders to know when the object had been untouched.

I think we need to do something similar, maybe in the ActiveCollisionContainer?

So when a collider is touched it gets added to the List and it's containing object is added to some internal state of currently touching objects, then when another collider is touched, it checks to see if that collider's containing object is already being touched, if it is then it doesn't run the touching routines again.

Then for untouch, maybe whenever a Collision/TriggerExit is called the collider's containing GameObject needs checking to see if any of it's colliders are being touched.

So if we keep a record of the collider touches on the parent (maybe a dictionary with the containing object as the key and a int count as the value?)

Then you can say, on Start Collision get my containing object and add it to the dictionary and increase the count by 1. If the count is 1 then it's a touch. If the count is greater than 1 then i'm just touching another collider. Then on untouch we look up the containing object, decrease the count by 1 and when the count is 0 we know it's fully untouched.

This may have a bunch of issues with it, one i can see is a potential race condition where i fully stop touching colliderA when i start touching colliderB

so it would put the untouch down to 0 (which does the untouch) and the touch to 1 again (which does the touch).

I'm not sure if Trigger/CollisionEnter is always called before Trigger/CollisionExit. If it is then that shouldn't be a problem because colliderB will be added to the count before colliderA is removed from the count. So when B is added there will be 2 colliders, then A gets removed and there is still 1 collider.

thestonefox:


Scripts/Data/Collection/GameObjectObservableCounter.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace VRTK.Core.Data.Collection
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using System;
+
+    /// <summary>
+    /// /// Allows observing counting the amount of attempts a <see cref="GameObject"/> is added or removed from a  <see cref="Dictionary{TKey, TValue}"/>.
```
bddckr:
```suggestion
    /// Allows counting the amount of attempts a <see cref="GameObject"/> is added or removed from a  <see cref="Dictionary{TKey, TValue}"/>.
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
+﻿namespace VRTK.Core.Data.Collection
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using System.Linq;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Allows observing counting the amount of attempts an element is added or removed from a <see cref="Dictionary{TKey, TValue}"/>.
```
bddckr:
```suggestion
    /// Allows counting the amount of attempts an element is added or removed from a <see cref="Dictionary{TKey, TValue}"/>.
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// </summary>
+    /// <typeparam name="TElement">The type of the elements to count.</typeparam>
+    /// <typeparam name="TEvent">The <see cref="UnityEvent"/> type to use.</typeparam>
+    public class ObservableCounter<TElement, TEvent> : MonoBehaviour where TEvent : UnityEvent<TElement>, new()
+    {
+        /// <summary>
+        /// Emitted when a searched element is found.
```
bddckr:
```suggestion
        /// Emitted when an element is added for the first time.
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </summary>
+        public TEvent ElementAdded = new TEvent();
+        /// <summary>
+        /// Emitted when a searched element is not found.
```
bddckr:
```suggestion
        /// Emitted when an element is removed completely.
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+        public virtual void IncreaseCount(TElement element)
+        {
+            if (!isActiveAndEnabled || element == null)
+            {
+                return;
+            }
+
+            if (!ElementsCounter.ContainsKey(element))
+            {
+                ElementsCounter.Add(element, 0);
```
bddckr: Remove the `if` and use this instead:
```suggestion
                ElementsCounter[element] = 0;
```
This indexer actually adds a key **or replaces** if the key already exists.

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// The elements being counted.
+        /// </summary>
+        public Dictionary<TElement, int> ElementsCounter
+        {
+            get;
+            protected set;
+        } = new Dictionary<TElement, int>();
+
+        /// <summary>
+        /// Increases the count on the given element.
```
bddckr:
```suggestion
        /// Increases the count of the given element.
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            ElementsCounter[element]++;
+            if (ElementsCounter[element] == 1)
+            {
+                ElementAdded?.Invoke(element);
+            }
+        }
+
+        /// <summary>
+        /// Decreases the count on the given element.
```
bddckr:
```suggestion
        /// Decreases the count of the given element.
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Removes the element from the counter.
+        /// </summary>
+        /// <param name="element">The element to clear.</param>
+        public virtual void RemoveFromCount(TElement element)
+        {
+            if (!isActiveAndEnabled || element == null || !ElementsCounter.ContainsKey(element))
```
bddckr: This dictionary lookup can be removed since `.Remove` actually returns a bool!
```suggestion
            if (!isActiveAndEnabled || element == null)
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                return;
+            }
+
+            ElementsCounter.Remove(element);
```
bddckr:
```suggestion
            if (ElementsCounter.Remove(element))
            {
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+            ElementRemoved?.Invoke(element);
```
bddckr:
```suggestion
                ElementRemoved?.Invoke(element);
            }
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// <summary>
+        /// Clears all elements from the counter.
+        /// </summary>
+        public virtual void Clear()
+        {
+            foreach (KeyValuePair<TElement, int> items in ElementsCounter.ToList())
```
bddckr:
```suggestion
            foreach (TElement element in ElementsCounter.Values.ToList())
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                RemoveFromCount(items.Key);
```
bddckr:
```suggestion
                RemoveFromCount(element);
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Retrieves the count value against a given element.
```
bddckr:
```suggestion
        /// How often an element was added to the counter without being removed.
```

Scripts/Data/Collection/ObservableCounter.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </summary>
+        /// <param name="element">The element to check.</param>
+        /// <returns>The count value for the given element.</returns>
```
bddckr:
```suggestion
        /// <returns>The count of the given element.</returns>
```

Scripts/Data/Collection/ObservableCounter.cs
thestonefox: That will always set the counter to 0 every time you call IncreaseCount, so count can only ever be 0 or 1

Scripts/Data/Collection/ObservableCounter.cs
thestonefox: I need to get the value out at that point to check if the value is less or equal zero to bomb out too.

If i do a remove there, then the following `ElementsCounter[element]--` will fail too

bddckr: This is the clear method which doesn't do any counter--. Are you confusing this method with the decrease one?

thestonefox: I was 😄

Scripts/Data/Collection/ObservableCounter.cs
thestonefox: Could this method not just be simplified to:

```
public virtual void RemoveFromCount(TElement element)
{
    if (!isActiveAndEnabled || element == null || !ElementsCounter.Remove(element))
    {
        return;
    }
    ElementRemoved?.Invoke(element);
}
```

if it's not active or enabled or the element is null or when i try and remove the element it doesn't find one to remove then bomb out early, otherwise invoke the method.

Scripts/Data/Collection/ObservableCounter.cs
thestonefox: `ElementsCounter.Keys.ToList` as I want the Element, not the int :)

Scripts/Data/Collection/ObservableCounter.cs
bddckr: 👍

Scripts/Data/Collection/ObservableCounter.cs
bddckr: 👍

PR #237, opened 2018-12-14
feat(Process): reimplement source target processing and object follow - fixes #236 - fixes #225

### LINKED ISSUE #225: The Follow Modifiers (Position/Rotation/Scale) don't act as expected when used in conjunction with each other
We need to have a session to review how these are working in unison as they don't produce the expected results

### LINKED ISSUE #236: Rework the ObjectFollower into two distinct components that have a single responsibility
At the moment the ObjectFollower script has two distinct functions of making target objects follow a source and a source object follow an active target.

It's more appropriate to have 2 distinct components that handle these functions that can extend a common base.

### Notes:

ObjectFollower becomes abstract class that extends SourceTargetProcessor

New simple class for holding relationship between a GameObject and a child offset

```
GameObjectWithOffsetContainer
  {
  GameObject entity
  GameObject childOffset
  }
  
  SetEntity(GameObject obj)
  ClearEntity()
  SetChildOffset(GameObject obj) (checks to see if the given obj is a child of entity)
  ClearChildOffset()
```

SourceTargetProcessor params
 * Source:Component - can take a GameObject or a GameObjectWithOffsetContainer
 * Targets:List<Component> - can take a collection of GameObjects or GameObjectWithOffsetContainers
 * Also needs setter/clear for targets list
   * SetTargetIndex(int)
   * SetTargetAtCurrentIndex(Component)
   * SetTarget(int, Component)
 * At the moment SourceTargetProcessor only offers methods for setting source/target by GameObject, but it should offer Component as well.

ObjectFollower params
 * FollowModifier - The modifier to process on.
 
Split out the existing logic into two distinct classes

----

 * SourceToActiveTargetFollower
   * the given source follows the first GameObject that is active in the target list
   * uses `ProcessFirstActiveComponent`
   * the source can be of type GameObject or GameObjectWithOffsetContainer
     * if the source is GameObject then it is passed to the FollowModifier as the target parameter
     * if the source is GameObjectWithOffsetContainer then
       * source.entity is passed to the FollowModifier as the target parameter
       * source.childOffset is passed to the FollowModifier as the offset parameter
   * the targets can be of type GameObject
     * the first active target is passed to the FollowModifier as the source parameter

----     

 * AllTargetsToSourceFollower
   * all of the given targets follow the given source GameObject
   * uses `ProcessAllComponents`
   * the source can be of type GameObject
     * the source is passed to the FollowModifier as the source parameter
   * the targets can be of type GameObject or GameObjectWithOffsetContainer
     * if the target is GameObject then it is passed to the FollowModifier as the target parameter
     * if the target is GameObjectWithOffsetContainer then
       * target[n].entity is passed to the FollowModifier as the target parameter
       * target[n].childOffset is passed to the FollowModifier as the offset parameter

thestonefox:
The SourceTargetProcessor has been re-written to be more generic and to
allow the ObjectFollower to be re-written so it is more cohesive.

The ObjectFollower now is consistent in having a collection of source
GameObjects that always are used for the source of data and a
collection of target GameObjects that always have the source data
applied to.

The source and target data can also accept a rule to determine which
source elements to use/ignore and which target elements to use/ignore
which makes it possible to replicate the original functionality of
the ObjectFollower, which was:

 * Targets follow source
 * Source follows active Target

But without the need to flip the meaning of the fields or require
enums with descriptions of confusing functionality.

Due to the changes to the SourceTargetProcessor, the
GameObjectStateMirror component also needed updating to bring it in
line with the changes.

A new ICollectionExtensions component has been added that provides
a simple helper for components that extend ICollection to retrieve
an index that is within the bounds of the collection and allows for
negative indices to make it possible to get elements starting from
the end of the collection.

The Follow modifiers have also been updated to no longer attempt
to set transform properties in local space as it makes no sense
when applying a follow operation. The offset must also be a child
of the target and always uses the offset local space in relation
to the target to offset against the source.

All of the prefabs that used a SourceTargetProcessor have been
updated to utilise the new mechanism.


Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
         {
             return (Vector2.Distance(a, b) <= tolerance);
         }
+
+        /// <summary>
+        /// Divides each component of the given <see cref="Vector2"/> against the given <see cref="float"/>.
+        /// </summary>
+        /// <param name="value">The value to divide against.</param>
+        /// <param name="vector">The <see cref="Vector2"/> components to divide with.</param>
```
bddckr:
```suggestion
        /// <param name="divisor">The components to divide with.</param>
```

Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
         {
             return (Vector2.Distance(a, b) <= tolerance);
         }
+
+        /// <summary>
+        /// Divides each component of the given <see cref="Vector2"/> against the given <see cref="float"/>.
+        /// </summary>
+        /// <param name="value">The value to divide against.</param>
+        /// <param name="vector">The <see cref="Vector2"/> components to divide with.</param>
+        /// <returns>The divided result.</returns>
```
bddckr:
```suggestion
        /// <returns>The quotient.</returns>
```

Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
...
             return (Vector2.Distance(a, b) <= tolerance);
         }
+
+        /// <summary>
+        /// Divides each component of the given <see cref="Vector2"/> against the given <see cref="float"/>.
+        /// </summary>
+        /// <param name="value">The value to divide against.</param>
+        /// <param name="vector">The <see cref="Vector2"/> components to divide with.</param>
+        /// <returns>The divided result.</returns>
+        public static Vector2 Divide(float value, Vector2 vector)
```
bddckr:
```suggestion
        public static Vector2 Divide(float dividend, Vector2 divisor)
```

Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
...
+        {
+            return new Vector2(value / vector.x, value / vector.y);
```
bddckr:
```suggestion
            return new Vector2(dividend / divisor.x, dividend / divisor.y);
```

Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
...
+        }
+        /// <summary>
+        /// Divides two <see cref="Vector2"/>s component-wise.
+        /// </summary>
+        /// <param name="a">The <see cref="Vector2"/> to divide against.</param>
+        /// <param name="b">The <see cref="Vector2"/> to divide with.</param>
+        /// <returns>The divided result.</returns>
+        public static Vector2 Divide(this Vector2 a, Vector2 b)
+        {
+            return Vector2.Scale(a, Divide(1, b));
```
bddckr:
```suggestion
            return Vector2.Scale(dividend, Divide(1f, divisor));
```

Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
...
+        {
+            return new Vector2(value / vector.x, value / vector.y);
+        }
+        /// <summary>
+        /// Divides two <see cref="Vector2"/>s component-wise.
+        /// </summary>
+        /// <param name="a">The <see cref="Vector2"/> to divide against.</param>
+        /// <param name="b">The <see cref="Vector2"/> to divide with.</param>
+        /// <returns>The divided result.</returns>
+        public static Vector2 Divide(this Vector2 a, Vector2 b)
```
bddckr:
```suggestion
        public static Vector2 Divide(this Vector2 dividend, Vector2 divisor)
```

Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
...
+        public static Vector2 Divide(float value, Vector2 vector)
+        {
+            return new Vector2(value / vector.x, value / vector.y);
+        }
+        /// <summary>
+        /// Divides two <see cref="Vector2"/>s component-wise.
+        /// </summary>
+        /// <param name="a">The <see cref="Vector2"/> to divide against.</param>
+        /// <param name="b">The <see cref="Vector2"/> to divide with.</param>
+        /// <returns>The divided result.</returns>
```
bddckr:
```suggestion
        /// <returns>The quotient.</returns>
```

Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
...
+        /// <param name="vector">The <see cref="Vector2"/> components to divide with.</param>
+        /// <returns>The divided result.</returns>
+        public static Vector2 Divide(float value, Vector2 vector)
+        {
+            return new Vector2(value / vector.x, value / vector.y);
+        }
+        /// <summary>
+        /// Divides two <see cref="Vector2"/>s component-wise.
+        /// </summary>
+        /// <param name="a">The <see cref="Vector2"/> to divide against.</param>
```
bddckr:
```suggestion
        /// <param name="dividend">The value to divide by each component.</param>
```

Scripts/Extension/Vector2Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector2 a, Vector2 b, float tolerance = flo
...
+        /// <param name="b">The <see cref="Vector2"/> to divide with.</param>
```
bddckr:
```suggestion
        /// <param name="divisor">The components to divide with.</param>
```

Scripts/Extension/Vector3Extensions.cs
```diff
@@ -18,5 +18,26 @@ public static bool ApproxEquals(this Vector3 a, Vector3 b, float tolerance = flo
         {
             return (Vector3.Distance(a, b) <= tolerance);
         }
+
```
bddckr: Same changes as done for the `Vector2` variants.

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+        public List<GameObject> targets = new List<GameObject>();
+        /// <summary>
+        /// Allows to optionally determine which targets should be processed based on the set rules.
+        /// </summary>
+        [Tooltip("Allows to optionally determine which targets should be processed based on the set rules.")]
+        public RuleContainer targetValidity;
+        #endregion
+
+        /// <summary>
+        /// The current sources collection index.
```
bddckr:
```suggestion
        /// The current <see cref="sources"/> collection index.
```

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </summary>
+        public int CurrentSourcesIndex
+        {
+            get { return _currentSourcesIndex; }
+            set { _currentSourcesIndex = sources.GetWrappedAndClampedIndex(value); }
+        }
+        private int _currentSourcesIndex;
+
+        /// <summary>
+        /// The current targets collection index.
```
bddckr:
```suggestion
        /// The current <see cref="targets"/> collection index.
```

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </summary>
+        public int CurrentTargetsIndex
+        {
+            get { return _currentTargetsIndex; }
+            set { _currentTargetsIndex = targets.GetWrappedAndClampedIndex(value); }
+        }
+        private int _currentTargetsIndex;
+
+        /// <inheritdoc />
+        public override void Process()
```
bddckr: Does it make sense to move this into the superclass instead? Subclasses can still override it, but it seems like a good default implementation of this method to have in `SourceTargetProcessor`.

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            sources.Remove(source);
+        }
+
+        /// <summary>
+        /// Sets the given source at the given index of the sources collection.
+        /// </summary>
+        /// <param name="index">The index to set the source at.</param>
+        /// <param name="source">The source to set.</param>
+        public virtual void SetSource(int index, GameObject source)
```
bddckr: `sources` is `public` and a mutable `List`, so is this needed?

Users can get access to the list themselves and mutate it that way and UnityEvents don't allow being wired up to a method taking more than one argument.

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            targets.Remove(target);
+        }
+
+        /// <summary>
+        /// Sets the given target at the given index of the targets collection.
+        /// </summary>
+        /// <param name="index">The index to set the target at.</param>
+        /// <param name="target">The target to set.</param>
+        public virtual void SetTarget(int index, GameObject target)
```
bddckr: Same as above: Potentially remove?

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// <summary>
+        /// Clears the targets collection.
+        /// </summary>
+        public virtual void ClearTargets()
+        {
+            targets.Clear();
+        }
+
+        /// <inheritdoc />
+        protected override void SetCurrentIndices(int sourceIndex, int targetIndex)
```
bddckr: Can this be moved into the superclass, too?

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            CurrentSourcesIndex = sourceIndex;
+            CurrentTargetsIndex = targetIndex;
+        }
+
+        /// <inheritdoc />
+        protected override bool IsSourceValid(GameObject source)
```
bddckr: Does it make sense for the superclass to have this validity stuff?

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            return (base.IsSourceValid(source) && sourceValidity.Accepts(source));
+        }
+
+        /// <inheritdoc />
+        protected override bool IsTargetValid(GameObject target)
```
bddckr: Same as above: Does it make sense for the superclass to have this validity stuff?

Scripts/Tracking/Follow/Modifier/Property/Scale/TransformScale.cs
```diff
@@ -22,13 +16,14 @@ public class TransformScale : PropertyModifier
...
-            if (useLocalScale)
+            if (offset == null)
             {
-                target.transform.localScale = (offset != null ? (source.transform.localScale - (offset.transform.localScale - target.transform.localScale)) : source.transform.localScale);
+                target.transform.SetGlobalScale(source.transform.lossyScale);
             }
             else
             {
-                target.transform.SetGlobalScale((offset != null ? (source.transform.lossyScale - (offset.transform.lossyScale - target.transform.lossyScale)) : source.transform.lossyScale));
+                Vector3 offsetMultiplier = Vector3Extensions.Divide(1, offset.transform.localScale);
```
bddckr:
```suggestion
```

Scripts/Tracking/Follow/Modifier/Property/Scale/TransformScale.cs
```diff
@@ -22,13 +16,14 @@ public class TransformScale : PropertyModifier
...
+                target.transform.SetGlobalScale(Vector3.Scale(source.transform.lossyScale, offsetMultiplier));
```
bddckr:
```suggestion
                target.transform.SetGlobalScale(source.transform.lossyScale.Divide(offset.transform.localScale));
```

Scripts/Tracking/Follow/ObjectFollower.cs
```diff
@@ -60,54 +61,25 @@ public class UnityEvent : UnityEvent<EventData>
...
         public FollowModifier followModifier;
+
         /// <summary>
-        /// Determines which component is to be applied to the other component.
-        /// </summary>
-        [Tooltip("Determines which component is to be applied to the other component.")]
-        public ModificationType modificationType;
-        /// <summary>
-        /// The mechanism of how to process the targets.
+        /// The current targetOffsets collection index.
```
bddckr:
```suggestion
        /// The current <see cref="targetOffsets"/> collection index.
```

Scripts/Tracking/Follow/ObjectFollower.cs
```diff
@@ -123,53 +95,81 @@ public enum ProcessTarget
...
         }
 
-        /// <inheritdoc />
-        protected override void ProcessComponent(Component source, Component target)
+        /// <summary>
+        /// Sets the given targetOffset at the given index of the targetOffsets collection.
+        /// </summary>
+        /// <param name="index">The index to set the targetOffset at.</param>
+        /// <param name="targetOffset">The targetOffset to set.</param>
+        public virtual void SetTargetOffset(int index, GameObject targetOffset)
```
bddckr: Same as above: Maybe remove this?

Tests/Editor/Extension/Vector2ExtensionsTest.cs
```diff
@@ -30,5 +30,23 @@ public void ApproxEqualsFalse()
             tolerance = 1f;
             Assert.IsFalse(a.ApproxEquals(b, tolerance));
         }
+
+        [Test]
+        public void DivideByFloat()
```
bddckr:
```suggestion
        public void DivideWithFloat()
```

Tests/Editor/Extension/Vector3ExtensionsTest.cs
```diff
@@ -30,5 +30,23 @@ public void ApproxEqualsFalse()
             tolerance = 1f;
             Assert.IsFalse(a.ApproxEquals(b, tolerance));
         }
+
+        [Test]
+        public void DivideByFloat()
```
bddckr:
```suggestion
        public void DivideWithFloat()
```

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
thestonefox: This just tidies up the `SetSourceAtCurrentIndex` method to simply call `SetSource`

But we could ditch this method and move the internals to `SetSourceAtCurrentIndex`

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
thestonefox: I don't think so because CurrentSourcesIndex has a setter that controls the `sources` list which is not defined in the superclass

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
thestonefox: The superclass calls the abstract method defined in the superclass

Scripts/Tracking/Follow/Modifier/Property/Scale/TransformScale.cs
thestonefox: I can't see what the suggested change is here because github is making it look all confusing.

are you saying instead of: 
```
Vector3 offsetMultiplier = Vector3Extensions.Divide(1, offset.transform.localScale);
target.transform.SetGlobalScale(Vector3.Scale(source.transform.lossyScale, offsetMultiplier));
```

to just do

```
target.transform.SetGlobalScale(source.transform.lossyScale.Divide(offset.transform.localScale));
```

If so, I think that causes issues where you can get divisions by zero. The previous way avoids divisions by zero (unless you actually put a 0 in for the local scale)

bddckr: My suggestion is to not have the variable above (see the suggestion deleting the whole line above this) because you introduced the extension method doing the exact same thing.

You already do `1f / vector` above, I didn't change any logic. I can see how that division can fail when dividing by zero, though.
I wouldn't worry about it as that would mean someone has set the local scale to zero, which is not really valid anyhow.

Scripts/Process/Component/GameObjectSourceTargetProcessor.cs
thestonefox: Only the `!isActiveAndEnabled` because the `sources` and `target` do not exist in the superclass.

thestonefox: But then doing the return in the superclass won't help because it will just exit early from the superclass method and keep running this method anyway.

bddckr: 👍

PR #250, opened 2019-01-01
feat(Query): add pointing query - resolves #52

### LINKED ISSUE #52: Determine if a given object is looking at a specified object
This will provide the basis to determine if the headset is looking at a specific controller.

bddckr:
A new component has been added that allows querying whether a
GameObject points at another GameObject. An example use case is
showing tooltips on a VR controller while it's viewed by the
headset.


Scripts/Tracking/Query/PointingQuery.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// <summary>
+    /// Determines whether a <see cref="GameObject"/> points at another <see cref="GameObject"/>.
+    /// </summary>
+    /// <remarks>
+    /// No physics checks are done and as such occlusion isn't part of the information gained by this component.
+    /// </remarks>
+    public class PointingQuery : MonoBehaviour, IProcessable
+    {
+        /// <summary>
+        /// The object that will be checked for pointing at <see cref="target"/>.
```
thestonefox: `The object used to as the origin to check if it is pointing towards the <see cref="target"/>`.

Scripts/Tracking/Query/PointingQuery.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        [Tooltip("The object that will be checked for pointing at target.")]
```
thestonefox: `The object that will be checked to see if the <see cref="source"/> is pointing at it.`

PR #251, opened 2019-01-03
docs(README.md): add getting started steps

thestonefox:
The Getting Started steps provide a step by step guide of how to
import VRTK.Unity.Core into an existing Unity3d project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,23 +11,45 @@
 
 ## 3<sup>rd</sup> Party SDK Support
 
- * [VRTK.Unity.SteamVR]
- * [VRTK.Unity.OculusUtilities]
+* [VRTK.Unity.SteamVR]
+* [VRTK.Unity.OculusUtilities]
 
 ## Introduction
 
-VRTK.Unity.Core is a core collection of design patterns for Unity
-that can be beneficial in VR development.
+VRTK.Unity.Core is a core collection of design patterns for [Unity3d] that can be beneficial in VR development.
+
+## Getting Started
+
+### Setting up a project
+
+* Using Unity3d 2018.1 or above create a new project using the 3D Template or open an existing project.
+* Update the project `Scripting Runtime Version` to `.NET 4.x Equivalent`.
+  * In Unity3d select `Main Menu -> Edit -> Project Settings -> Player` to open the `Player Settings` inspector.
+  * In the `PlayerSettings` inspector panel expand `Other Settings`.
+  * Change the `Scripting Runtime Version` to `.NET 4.x Equivalent`.
+  * Unity3d will now restart in the supported scripting runtime.
+* If the project supports Virtual Reality then ensure `Virtual Reality Supported` is checked.
+  * In Unity3d select `Main Menu -> Edit -> Project Settings -> Player` to open the `Player Settings` inspector.
+  * In the `PlayerSettings` inspector panel expand `XR Settings`.
+  * In `XR Settings` ensure the `Virtual Reality Supported` option is checked.
+
+### Cloning the VRTK.Unity.Core repo into a project
+
+* Navigate to the `Assets/` directory of your project.
+* Git clone this repo into the `Assets/` directory:
+  * `git clone https://github.com/ExtendRealityLtd/VRTK.Unity.Core.git` - [How To Clone A Repo]
+* Wait for Unity3d to finish importing the cloned files.
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,18 +11,38 @@
...
 
-VRTK.Unity.Core is a core collection of design patterns for Unity
-that can be beneficial in VR development.
+VRTK.Unity.Core is a core collection of design patterns for [Unity3d] that can be beneficial in VR development.
+
+## Getting Started
+
+### Setting up a project
+
+  * Create a new project using the 3D Template or open an existing project in [Unity3d] 2018.1 or above.
```
bddckr:
```suggestion
  * Using [Unity3d] 2018.1 or above create a new project using the 3D Template or open an existing project.
```

README.md
```diff
@@ -11,18 +11,38 @@
...
+  * Update the project `Scripting Runtime Version` to `.NET 4.x Equivalent`.
```
bddckr: Starting with 2018.3 this is actually the default and thus we'd need no instructions for this at all... I don't think we want to decide on only supporting 2018.3 already, though. 🤔

The project settings are also in their own standalone window starting with that version, so those instructions may not match perfectly anymore.

README.md
```diff
@@ -11,18 +11,38 @@
...
- * [VRTK.Unity.SteamVR]
- * [VRTK.Unity.OculusUtilities]
+  * [VRTK.Unity.SteamVR]
+  * [VRTK.Unity.OculusUtilities]
 
 ## Introduction
 
-VRTK.Unity.Core is a core collection of design patterns for Unity
-that can be beneficial in VR development.
+VRTK.Unity.Core is a core collection of design patterns for [Unity3d] that can be beneficial in VR development.
```
bddckr: I'm a friend of following what Wikipedia and the likes do: Only the first occurrence of a term should be linked.

This would mean we remove the hyperlink to Unity here and anywhere below as the term is mentioned right at the top already.

README.md
```diff
@@ -11,18 +11,38 @@
...
+
+### Setting up a project
+
+  * Create a new project using the 3D Template or open an existing project in [Unity3d] 2018.1 or above.
+  * Update the project `Scripting Runtime Version` to `.NET 4.x Equivalent`.
+    * In [Unity3d] select `Main Menu -> Edit -> Project Settings -> Player` to open the `Player Settings` inspector.
+    * In the `PlayerSettings` inspector panel expand `Other Settings`.
+    * Change the `Scripting Runtime Version` to `.NET 4.x Equivalent`.
+    * [Unity3d] will now restart in the supported scripting runtime.
+  * If the project is for Virtual Reality then ensure `Virtual Reality Supported` is checked.
```
bddckr:
```suggestion
  * If the project supports Virtual Reality then ensure `Virtual Reality Supported` is checked.
```

README.md
bddckr: Again: Do we need to indent here at all?

README.md
```diff
@@ -11,18 +11,38 @@
...
+    * In [Unity3d] select `Main Menu -> Edit -> Project Settings -> Player` to open the `Player Settings` inspector.
+    * In the `PlayerSettings` inspector panel expand `XR Settings`.
+    * In `XR Settings` ensure the `Virtual Reality Supported` option is checked.
+
+### Cloning the VRTK.Unity.Core repo into a project
+
+  * Navigate to the `Assets/` directory of your project.
```
bddckr: Again: Do we need to indent here at all?

README.md
```diff
@@ -11,18 +11,38 @@
...
+  * Git clone this repo into the `Assets/` directory:
+    * `git clone https://github.com/ExtendRealityLtd/VRTK.Unity.Core.git`
```
bddckr:
```suggestion
    * [`git clone https://github.com/ExtendRealityLtd/VRTK.Unity.Core.git`][Clone Instructions]
```

With the matching
```markdown
[Clone Instructions]: https://help.github.com/articles/cloning-a-repository/
```
at the bottom of the file (in the correct place of the existing links).

README.md
thestonefox: These instructions are based on 2018.1 as there's no conclusion on the 2018.3 stuff yet, so if that becomes concrete these can change in the future.

README.md
bddckr: This section is still missing the removal of the indentation.

PR #253, opened 2019-01-07
Feat/2018.3 support

thestonefox:


README.md
```diff
@@ -9,29 +9,33 @@
...
-  * Unity3d will now restart in the supported scripting runtime.
-* If the project supports Virtual Reality then ensure `Virtual Reality Supported` is checked.
-  * In Unity3d select `Main Menu -> Edit -> Project Settings -> Player` to open the `Player Settings` inspector.
-  * In the `PlayerSettings` inspector panel expand `XR Settings`.
-  * In `XR Settings` ensure the `Virtual Reality Supported` option is checked.
+* Using Unity3d 2018.3 (or above) create a new project using the 3D Template or open an existing project.
+* If the project supports Virtual Reality:
+  * Ensure `Virtual Reality Supported` is checked.
+    * In Unity3d select `Main Menu -> Edit -> Project Settings` to open the `Project Settings` inspector.
+    * Select `Player` from the left hand menu in the `Project Settings` inspector.
```
bddckr:
```suggestion
    * Select `Player` from the left hand menu in the `Project Settings` window.
```
Starting with .3 it's (finally) a window.

README.md
```diff
@@ -9,29 +9,33 @@
...
+    * In the `Player` inspector panel expand `XR Settings`.
+    * In `XR Settings` ensure the `Virtual Reality Supported` option is checked.
+  * Install the appropriate support package.
```
bddckr:
```suggestion
  * Ensure the appropriate support package is installed.
```
Unity _should_ be adding the package automatically once enabled, thus these instructions just ensure that, they don't necessarily add the package as they may already exist.

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        [Tooltip("The duration to pulse the device for.")]
+        public float duration = 0.005f;
+
+        /// <summary>
+        /// The haptic capabilities of the specified <see cref="XRNode"/>.
+        /// </summary>
+        protected HapticCapabilities caps = new HapticCapabilities();
+
+        /// <inheritdoc />
+        public override bool IsActive()
```
bddckr: Remove this method as it's the same as the base implementation.

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
+﻿namespace VRTK.Core.Haptics
+{
+    using UnityEngine;
+    using UnityEngine.XR;
+
+    /// <summary>
+    /// Creates a timed haptic pulse on a <see cref="XRNode"/> supported haptic device.
```
bddckr:
```suggestion
    /// Creates a timed haptic pulse on a <see cref="XRNode"/>.
```
A "device" is a different thing in the UnityEngine.XR namespace, thus we should only talk about "nodes" here to prevent confusion.

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// </summary>
+    public class XRNodeHapticPulse : HapticPulse
+    {
+        /// <summary>
+        /// The device to pulse.
```
bddckr:
```suggestion
        /// The node to pulse.
```
Again: Devices != nodes according to Unity.

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        [Tooltip("The device to pulse.")]
```
bddckr:
```suggestion
        [Tooltip("The node to pulse.")]
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        public XRNode device = XRNode.LeftHand;
```
bddckr:
```suggestion
        public XRNode node = XRNode.LeftHand;
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        /// The duration to pulse the device for.
```
bddckr:
```suggestion
        /// The duration to pulse <see cref="node"/> for.
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        [Tooltip("The duration to pulse the device for.")]
```
bddckr:
```suggestion
        [Tooltip("The duration to pulse node for.")]
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        public float duration = 0.005f;
+
+        /// <summary>
+        /// The haptic capabilities of the specified <see cref="XRNode"/>.
```
bddckr:
```suggestion
        /// The haptic capabilities of <see cref="node"/>.
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        protected HapticCapabilities caps = new HapticCapabilities();
```
bddckr:
```suggestion
        protected HapticCapabilities capabilities = new HapticCapabilities();
```
We rarely use abbreviations.

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <inheritdoc />
+        public override bool IsActive()
+        {
+            return base.IsActive();
+        }
+
+        /// <inheritdoc />
+        protected override void DoBegin()
+        {
+            Pulse(Intensity, duration, device);
```
bddckr:
```suggestion
            Pulse(Intensity, duration);
```
The `node` is an instance variable, thus we don't need to pass it in.

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        /// <inheritdoc />
+        protected override void DoCancel()
+        {
+            Pulse(0, 0, device);
```
bddckr:
```suggestion
            Pulse(0f, 0f);
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        /// <summary>
+        /// Sends a pulse to the current device.
```
bddckr:
```suggestion
        /// Sends a pulse to <see cref="node"/>.
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        /// <param name="intensity">The intensity to pulse with.</param>
+        /// <param name="duration">The duration to pulse for.</param>
+        /// <param name="node">The device to pulse on.</param>
```
bddckr:
```suggestion
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        protected virtual void Pulse(float intensity, float duration, XRNode node)
```
bddckr:
```suggestion
        protected virtual void Pulse(float intensity, float duration)
```

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            if (!InputDevices.GetDeviceAtXRNode(node).TryGetHapticCapabilities(out caps))
```
bddckr: Put the output of `InputDevices.GetDeviceAtXRNode(node)` into a local variable before this `if` and reuse it in the whole method, that way we only call this method once.

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        /// <param name="duration">The duration to pulse for.</param>
+        /// <returns>The buffer array containing the pulse data.</returns>
+        protected virtual byte[] GeneratePulseBuffer(float duration)
+        {
+            int clipCount = (int)(caps.bufferFrequencyHz * duration);
+            byte[] clip = new byte[clipCount];
+            for (int i = 0; i < clipCount; i++)
+            {
+                clip[i] = byte.MaxValue;
```
bddckr: This sets a buffer that is at full intensity at all times, see [these tests](https://github.com/Unity-Technologies/XRInputTests/tree/2018.3/Assets/Tests/HapticsTest/Scripts/ClipTests).

This is what we want in this case I think, so this is more of a comment than a review comment. Sadly we can't offer specific API to use this as the buffer support isn't guaranteed for everything and it will be hard to create a fallback that does the same by taking the user's buffer and then trying to call the impulse API to represent that buffer (if pulses are supported by the node that is).

Scripts/Haptics/XRNodeHapticPulse.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        /// Generates a pulse buffer array.
+        /// </summary>
+        /// <param name="duration">The duration to pulse for.</param>
+        /// <returns>The buffer array containing the pulse data.</returns>
+        protected virtual byte[] GeneratePulseBuffer(float duration)
+        {
+            int clipCount = (int)(caps.bufferFrequencyHz * duration);
+            byte[] clip = new byte[clipCount];
+            for (int i = 0; i < clipCount; i++)
```
bddckr: Do we do `i` or `index`? I can't remember. 😅

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace VRTK.Core.Tracking.Velocity
+{
+    using System.Collections.Generic;
+    using UnityEngine;
+    using UnityEngine.XR;
+
+    /// <summary>
+    /// Retrieves the velocity estimations for the specified <see cref="XRNode"/> device.
```
bddckr:
```suggestion
    /// Retrieves the velocity estimations for a <see cref="XRNode"/>.
```

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// </summary>
+    public class XRNodeVelocityEstimator : VelocityTracker
+    {
+        /// <summary>
+        /// The device to query velocity estimations for.
```
bddckr:
```suggestion
        /// The node to query velocity estimations for.
```

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// </summary>
+        [Tooltip("The device to query velocity estimations for.")]
```
bddckr:
```suggestion
        [Tooltip("The node to query velocity estimations for.")]
```

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,59 @@
...
+        public XRNode device = XRNode.LeftHand;
```
bddckr:
```suggestion
        public XRNode node = XRNode.LeftHand;
```

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <inheritdoc />
+        public override bool IsActive()
+        {
+            return isActiveAndEnabled;
```
bddckr: Move this into the base class instead (and update other subclasses of that if necessary).

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <inheritdoc />
+        protected override Vector3 DoGetAngularVelocity()
+        {
+            Vector3 result = Vector3.zero;
+            GetNodeState().TryGetAngularVelocity(out result);
+            return result;
+        }
+
+        /// <summary>
+        /// Gets the state of the node associated with the specified <see cref="XRNode"/> device.
```
bddckr:
```suggestion
        /// Gets the state of <see cref="node"/>.
```

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// </summary>
+        /// <returns>The associated node state.</returns>
+        protected virtual XRNodeState GetNodeState()
+        {
+            List<XRNodeState> nodesStates = new List<XRNodeState>();
```
bddckr: Make this a `protected` instance variable that we can reuse. This makes this API usage almost always allocation-free.

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,59 @@
...
+            InputTracking.GetNodeStates(nodesStates);
```
bddckr:
```suggestion
            return InputTracking.GetNodeStates(nodesStates).FirstOrDefault(state => state.nodeType == node);
```
Then remove the rest of this method.

bddckr:
Once merged let's retire the SDK repos (by making them private).

Scripts/Haptics/XRNodeHapticPulse.cs
thestonefox: looks like `i` in most places and `index` in some.

Personally, in for loops i don't mind a single `i` for the index counter.

But I guess we need to be consistent over the entire code base.

Whatever we decide here, we should apply across the whole codebase soon too.

Scripts/Haptics/XRNodeHapticPulse.cs
thestonefox: Intensity is a normalized value between 0 - 1. Can't we just multiple `byte.MaxValue` by `Intensity`? then if it's 1 you get the max value.?

`clip[i] = (byte)(byte.MaxValue * Intensity);`

Scripts/Haptics/XRNodeHapticPulse.cs
bddckr: 👍 Consistency!

I prefer `index` because if you have nested loops you end up with `i`, `j` and so on which is mainly what I dislike, one also often falls into the trap of doing that even if the index isn't just an index (e.g. math expressions). That's why I settled on always typing it out and then I end up with names like `clipIndex` and `bufferIndex` in the hypothetical example of multiple buffers for multiple clips being generated. Or in math you get reminded to give descriptive names in case the actual index _value_ is of interest - `y` instead of `j` etc.

thestonefox: Ok I'll raise a ticket to update everything so it's no longer `i`

Scripts/Haptics/XRNodeHapticPulse.cs
bddckr: Oh perfect - I totally didn't come up with that because the `intensity` is not yet passed into this method as a parameter. Definitely do this (and remember to add the XML docs for this additional parameter - I often forget that one 😅).

Scripts/Haptics/XRNodeHapticPulse.cs
bddckr: Feel free to just mark this comment thread as resolved - we can change this `i` to `index` once we do that ticket then.

thestonefox: I may as well change this one now as i'm rebasing anyway :)

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
thestonefox: I'll do this in a separate commit as it will affect many other velocity trackers and doesn't feel right doing it here.

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
thestonefox: That doesn't work as `GetNodeStates` returns void, it's basically piggy backing on nodeStates being a reference parameter. it's so terrabad I can't begin to comprehend the reason why it works that way!

anyway this works

```
nodesStates.Clear();
InputTracking.GetNodeStates(nodesStates);
return nodesStates.FirstOrDefault(state => state.nodeType == node);
```

(i'm just clearing the nodeStates list first as it's now an instance variable and I dunno how clean it is so a quick clear shouldn't hurt)

bddckr: According to the [docs](https://docs.unity3d.com/ScriptReference/XR.InputTracking.GetNodeStates.html) Unity does the same so we shouldn't call `Clear` to prevent that unnecessary additional call.

Scripts/Haptics/XRNodeHapticPulse.cs
bddckr: This still needs to be done.

Scripts/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// The node to query velocity estimations for.
+        /// </summary>
+        [Tooltip("The node to query velocity estimations for.")]
+        public XRNode node = XRNode.LeftHand;
+
+        /// <summary>
+        /// A collection of node states.
+        /// </summary>
+        protected List<XRNodeState> nodesStates = new List<XRNodeState>();
```
bddckr:
```suggestion
        protected readonly List<XRNodeState> nodesStates = new List<XRNodeState>();
```

PR #259, opened 2019-01-21
Fix/for index counters - fixes #254

### LINKED ISSUE #254: Replace all instances of `for` index counters from being single letter variables
We have an inconsistency of some `for` loops using `int i` and others using `int index`

it's our preference to never use abbreviations to avoid confusion, so any instance of `int i` should be replaced with `int index` or renaming `index` to a more appropriate name e.g. `int clipIndex`

thestonefox:


.github/CONTRIBUTING.md
```diff
@@ -70,50 +70,121 @@ and use American spellings to denote classes, methods, fields, etc.
...
+    {
+    }
+  }
+  ```
+
+  > **Correct:**
+  ```
+  namespace X.Y.Z
+  {
+    using System;
```
bddckr: Add a newline after this to ensure folks understand that part of our conventions?

.github/CONTRIBUTING.md
```diff
@@ -70,50 +70,121 @@ and use American spellings to denote classes, methods, fields, etc.
...
-    {
-    }
+    for (int clipIndex = 0; clipIndex < 10; clipIndex++) {}
   }
   ```
 
+Class methods and parameters should always denote their accessibility
+level using the `public` `protected` `private` keywords. Methods should
+also always be defined as virtual in concrete classes to allow
+overriding.
```
bddckr:
```suggestion
overriding. `protected` is favored until a good reason to make it `private` is found. This ensures the type is open for extension where possible.
```

Scripts/Utility/BezierCurveGenerator.cs
```diff
@@ -11,42 +11,47 @@ public static class BezierCurveGenerator
         /// <summary>
         /// Generates points on a bezier curve.
         /// </summary>
-        /// <param name="count">The number of points to generate.</param>
+        /// <param name="pointsToGenerate">The number of points to generate.</param>
```
bddckr: `pointsToGenerate` sounds like one hands over actual points. `pointsCount` perhaps?

PR #277, opened 2019-02-11
feat(Data): add zero value shortcut to FloatRange

thestonefox:
The default constructor of the FloatRange creates a FloatRange with a
float minimum and float maximum value, but on occasion it is useful to
set the default to zero (0f, 0f) without having to manually create the
constructor so a shorthand static FloatRange.Zero has been added.

### DIFF
--- a/Runtime/Data/Type/FloatRange.cs
+++ b/Runtime/Data/Type/FloatRange.cs
@@ -22,6 +22,11 @@
         public float maximum;
 
         /// <summary>
+        /// Shorthand for writing <c>FloatRange(0f, 0f)</c>.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


Runtime/Data/Type/FloatRange.cs
```diff
@@ -21,6 +21,11 @@ public class FloatRange
         [DocumentedByXml]
         public float maximum;
 
+        /// <summary>
+        /// Shorthand for writing FloatRange(0f, 0f).
```
bddckr:
```suggestion
        /// Shorthand for writing <c>FloatRange(0f, 0f)</c>.
```
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/code-inline

PR #278, opened 2019-02-11
fix(Modification): add threshold to destination equality check on lerp

thestonefox:
The TransformPropertyApplier could get into a situation where the
elapsed time would be a significantly small enough number away from the
transition duration to prevent the lerp from actually finishing and
would continue to update the properties of the transform by a very tiny
amount.

This fix adds a new threshold to check that the start and destination
properties of the transform scale, rotation and position are considered
equal to force the transition coroutine to finish and set the
properties to their final destination values.

### DIFF
--- a/Runtime/Tracking/Modification/TransformPropertyApplier.cs
+++ b/Runtime/Tracking/Modification/TransformPropertyApplier.cs
@@ -9,10 +9,10 @@
     using Malimbe.PropertySerializationAttribute;
     using Malimbe.PropertyValidationMethod;
     using Malimbe.XmlDocumentationAttribute;
+    using Zinnia.Extension;
     using Zinnia.Data.Attribute;
     using Zinnia.Data.Enum;
     using Zinnia.Data.Type;
-    using Zinnia.Extension;
 
     /// <summary>
     /// Applies the transform properties from a given source <see cref="Transform"/> onto the given target <see cref="Transform"/>.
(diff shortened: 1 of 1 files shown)


Runtime/Tracking/Modification/TransformPropertyApplier.cs
```diff
@@ -335,18 +340,35 @@ protected virtual IEnumerator TransitionTransform(TransformData source, Transfor
...
+        /// <summary>
+        /// Determines whether the start properties equal the destination properties.
+        /// </summary>
+        /// <param name="startPosition">The initial position of the <see cref="Transform"/>.</param>
+        /// <param name="destinationPosition">The final position for the <see cref="Transform"/>.</param>
+        /// <param name="startRotation">The initial rotation of the <see cref="Transform"/>.</param>
+        /// <param name="destinationRotation">The final rotation of the <see cref="Transform"/>.</param>
+        /// <param name="startScale">The initial scale of the <see cref="Transform"/>.</param>
+        /// <param name="destinationScale">The final scale of the <see cref="Transform"/>.</param>
+        /// <returns>Whether the start propeties equal the destination properties.</returns>
```
bddckr:
```suggestion
        /// <returns>Whether the start properties equal the destination properties.</returns>
```

Runtime/Tracking/Modification/TransformPropertyApplier.cs
```diff
@@ -335,18 +340,35 @@ protected virtual IEnumerator TransitionTransform(TransformData source, Transfor
...
+        protected virtual bool StartEqualsDestination(Vector3 startPosition, Vector3 destinationPosition, Quaternion startRotation, Quaternion destinationRotation, Vector3 startScale, Vector3 destinationScale)
```
bddckr:
```suggestion
        protected virtual bool ArePropertiesEqual(Vector3 startPosition, Vector3 destinationPosition, Quaternion startRotation, Quaternion destinationRotation, Vector3 startScale, Vector3 destinationScale)
```
In C# the convention seems to be offering special advice for `bool` returning members, including methods: They should answer a question.

PR #280, opened 2019-02-11
feat(Follow): rotate target by following position change of source

thestonefox:
The TransformPositionDifferenceRotation rotates the target by following
the position change of the source object allowing a pseudo rotation of
an object to mimic spinning an object around an axis based on the
position of the controlling source objects.

### DIFF
--- a/Runtime/Tracking/Follow/Modifier/Property/Rotation/TransformPositionDifferenceRotation.cs
+++ b/Runtime/Tracking/Follow/Modifier/Property/Rotation/TransformPositionDifferenceRotation.cs
@@ -1,0 +1,121 @@
+﻿namespace Zinnia.Tracking.Follow.Modifier.Property.Rotation
+{
+    using UnityEngine;
+    using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertyValidationMethod;
+    using Malimbe.PropertySerializationAttribute;
+    using Zinnia.Extension;
+    using Zinnia.Data.Type;
+
+    /// <summary>
+    /// Updates the transform rotation of the target to match the difference in position of the source.
+    /// </summary>
+    public class TransformPositionDifferenceRotation : PropertyModifier
+    {
+        /// <summary>
+        /// The drag applied to the rotation to slow it down.
+        /// </summary>
+        [Serialized, Validated]
+        [field: DocumentedByXml]
+        public float AngularDrag { get; set; } = 1f;
+
+        /// <summary>
+        /// Determines which axes to rotate.
+        /// </summary>
+        [Serialized, Validated]
+        [field: DocumentedByXml]
+        public Vector3State FollowOnAxis { get; set; } = Vector3State.True;
+
+        /// <summary>
+        /// The current angular velocity the rotation is applying to the target.
+        /// </summary>
+        public Vector3 AngularVelocity { get; protected set; }
+
+        /// <summary>
+        /// The previous source world position.
+        /// </summary>
+        protected Vector3? previousSourcePosition;
+
+        /// <summary>
+        /// Resets the state of the source previous position.
+        /// </summary>
+        public virtual void ResetPreviousState()
+        {
+            previousSourcePosition = null;
+        }
+
+        /// <summary>
+        /// Rotates the target based on the position difference of the source.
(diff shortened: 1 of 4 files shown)


Runtime/Tracking/Follow/Modifier/Property/Rotation/TransformPositionDifferenceRotation.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// </summary>
+        [Serialized, Validated]
+        [field: DocumentedByXml]
+        public Vector3State FollowOnAxis { get; set; } = Vector3State.True;
+
+        /// <summary>
+        /// The current angular velocity the rotation is applying to the target.
+        /// </summary>
+        public Vector3 AngularVelocity { get; protected set; }
+        /// <summary>
```
bddckr:
```suggestion

        /// <summary>
```
I think we always "group" members by accessibility using newlines.

Runtime/Tracking/Follow/Modifier/Property/Rotation/TransformPositionDifferenceRotation.cs
```diff
@@ -0,0 +1,98 @@
...
+    using Zinnia.Extension;
+    using Zinnia.Data.Type;
+
+    /// <summary>
+    /// Updates the transform rotation of the target to match the difference in position of the source.
+    /// </summary>
+    public class TransformPositionDifferenceRotation : PropertyModifier
+    {
+        /// <summary>
+        /// The drag applied to the rotation.
```
bddckr:
```suggestion
        /// The drag applied to the rotation to slow it down.
```
Just to prevent the user to lookup what "drag" is.

Runtime/Tracking/Follow/Modifier/Property/Rotation/TransformPositionDifferenceRotation.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <param name="originDirection">The direction representing the axis.</param>
+        /// <param name="originPoint">The angle centre.</param>
+        /// <param name="pointA">The first point to calculate the angle from.</param>
+        /// <param name="pointB">The second point to calculate the angle to.</param>
+        /// <returns>The angle in degrees between the two points.</returns>
+        protected virtual float CalculateAngle(Vector3 originDirection, Vector3 originPoint, Vector3 pointA, Vector3 pointB)
+        {
+            Vector3 heading = pointB - originPoint;
+            Vector3 sideA = pointA - originPoint;
+            if (heading.magnitude.ApproxEquals(0f))
```
bddckr: Save the result of `heading.magnitude` into a local variable and reuse it everywhere in this method - this property is calculated on the fly and does some math we don't want to be done multiple times.

Runtime/Tracking/Follow/Modifier/Property/Rotation/TransformPositionDifferenceRotation.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (previousSourcePosition == null)
+            {
+                previousSourcePosition = sourcePosition;
+            }
+
+            float xDegree = FollowOnAxis.xState ? CalculateAngle(target.transform.right, targetPosition, (Vector3)previousSourcePosition, sourcePosition) : 0f;
+            float yDegree = FollowOnAxis.yState ? CalculateAngle(target.transform.up, targetPosition, (Vector3)previousSourcePosition, sourcePosition) : 0f;
+            float zDegree = FollowOnAxis.zState ? CalculateAngle(target.transform.forward, targetPosition, (Vector3)previousSourcePosition, sourcePosition) : 0f;
+
+            AngularVelocity = new Vector3(xDegree, yDegree, zDegree) * (1f / AngularDrag);
```
bddckr: I suggest to move the `* (1f / AngularDrag)` part of this into its own method. This way subclasses can change the way the drag works, without having to replicated this whole method.

The method could be
```c#
protected virtual Vector3 ApplyDrag(Vector3 angularVelocity)
{
    return angularVelocity * (1f / AngularDrag);
}
```
perhaps (plus XML documentation comments obviously).

PR #290, opened 2019-02-22
Perf/garbage collected allocations

bddckr:


Runtime/Tracking/Collision/CollisionNotifier.cs
```diff
@@ -167,8 +167,8 @@ public enum CollisionTypes
         /// <returns><see langword="true"/> if events should be emitted.</returns>
         protected virtual bool CanEmit(EventData data)
         {
-            return (data.isTrigger && emittedTypes.HasFlag(CollisionTypes.Trigger)
-                    || !data.isTrigger && emittedTypes.HasFlag(CollisionTypes.Collision))
+            return (data.isTrigger && (emittedTypes & CollisionTypes.Trigger) != 0
```
thestonefox: should we be explicit here and do `emittedTypes != null`?

I know it's not "needed" but in some places you do need to be explicit (cos Unity) and at least the code then looks consistent

Runtime/Data/Collection/ObservableCounter.cs
```diff
@@ -45,14 +45,13 @@ public virtual void IncreaseCount(TElement element)
                 return;
             }
 
-            if (!ElementsCounter.ContainsKey(element))
+            if (ElementsCounter.ContainsKey(element))
             {
-                ElementsCounter.Add(element, 0);
+                ElementsCounter[element]++;
```
thestonefox: Doesn't this fail if `ElementsCounter[element]` doesn't already exist in the dictionary because it is trying to do an operation on something that doesn't exist by the time ElementsCounter adds `element`?

Runtime/Tracking/Collision/CollisionNotifier.cs
bddckr: I agree, but in this case this is doing a bitwise check on `emittedTypes` - the actual expression is `(emittedTypes & CollisionTypes.Trigger) != 0`. It's not doing a logical AND, it's a bitwise AND :)

bddckr:
Just pushed a tiny additional change to with `perf(general): reduce heap allocations by removing LINQ usages`: I removed some LINQ usage I missed in `SurfaceLocator`.

bddckr:
And another usage of LINQ in `ObjectPointer`. This should be it now. 😄

Runtime/Tracking/Collision/CollisionNotifier.cs
thestonefox: Oh yeah I didn't see the single & 😄

PR #293, opened 2019-02-24
Perf/improvements

bddckr:


thestonefox:
Everything seems to be ok and the tests pass. Obviously testing in VRTK is the real test if anything crops up but I say we wing it and suck it and see.

PR #299, opened 2019-02-27
feat(Collision): add ability to ignore collisions between GameObjects

thestonefox:
The CollisionIgnorer component takes a list of source GameObjects
and a list of target GameObjects and ignores all source colliders
from colliding with all target colliders.

The CollisionIgnorer ignores the collisions on enable and restores
the collisions on disable.

There is no direct access to the lists via code and new items must
be added via methods and removed via methods.

Adding new items to the list at runtime via the inspector yields no
changes to the collision detection.

### DIFF
--- a/Runtime/Tracking/Collision/CollisionIgnorer.cs
+++ b/Runtime/Tracking/Collision/CollisionIgnorer.cs
@@ -1,0 +1,222 @@
+﻿namespace Zinnia.Tracking.Collision
+{
+    using UnityEngine;
+    using System.Collections.Generic;
+    using Malimbe.MemberChangeMethod;
+    using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
+    using Zinnia.Data.Collection;
+
+    /// <summary>
+    /// Ignores the collisions between the source <see cref="GameObject"/> colliders and the target <see cref="GameObject"/> colliders.
+    /// </summary>
+    public class CollisionIgnorer : MonoBehaviour
+    {
+        /// <summary>
+        /// The sources to ignore colliders from.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public GameObjectObservableList Sources { get; set; }
+        /// <summary>
+        /// The targets to ignore colliders with.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public GameObjectObservableList Targets { get; set; }
+
+        /// <summary>
+        /// A reused instance to store the <see cref="Collider"/> collection belonging to the <see cref="Sources"/>.
+        /// </summary>
+        List<Collider> sourceColliders = new List<Collider>();
+        /// <summary>
+        /// A reused instance to store the <see cref="Collider"/> collection belonging to the <see cref="Targets"/>.
+        /// </summary>
+        List<Collider> targetColliders = new List<Collider>();
+
+        protected virtual void OnEnable()
+        {
+            RegisterSourceListeners();
+            RegisterTargetListeners();
+            ToggleCollisions(true);
+        }
+
+        protected virtual void OnDisable()
+        {
+            UnregisterSourceListeners();
+            UnregisterTargetListeners();
+            ToggleCollisions(false);
+        }
+
+        /// <summary>
+        /// Registers the listeners for elements that are added or removed from <see cref="Sources"/>.
+        /// </summary>
+        protected virtual void RegisterSourceListeners()
+        {
+            Sources.ElementAdded.AddListener(OnSourceAdded);
+            Sources.ElementRemoved.AddListener(OnSourceRemoved);
+        }
+
+        /// <summary>
+        /// Unregisters the listeners for elements that are added or removed from <see cref="Sources"/>.
+        /// </summary>
+        protected virtual void UnregisterSourceListeners()
+        {
+            Sources.ElementAdded.RemoveListener(OnSourceAdded);
+            Sources.ElementRemoved.RemoveListener(OnSourceRemoved);
+        }
+
+        /// <summary>
+        /// Registers the listeners for elements that are added or removed from <see cref="Targets"/>.
+        /// </summary>
+        protected virtual void RegisterTargetListeners()
+        {
+            Targets.ElementAdded.AddListener(OnTargetAdded);
+            Targets.ElementRemoved.AddListener(OnTargetRemoved);
+        }
+
+        /// <summary>
+        /// Unregisters the listeners for elements that are added or removed from <see cref="Targets"/>.
+        /// </summary>
+        protected virtual void UnregisterTargetListeners()
+        {
+            Targets.ElementAdded.RemoveListener(OnTargetAdded);
+            Targets.ElementRemoved.RemoveListener(OnTargetRemoved);
+        }
+
+        /// <summary>
+        /// Responds to a <see cref="GameObject"/> being added to <see cref="Sources"/> and ignores all collisions against <see cref="Targets"/>.
+        /// </summary>
+        /// <param name="source">The source to ignore collisions from.</param>
+        protected virtual void OnSourceAdded(GameObject source)
+        {
+            ToggleCollisions(source, Sources, Targets, true);
+        }
+
+        /// <summary>
+        /// Responds to a <see cref="GameObject"/> being removed from <see cref="Sources"/> and resumes all collisions against <see cref="Targets"/>.
+        /// </summary>
+        /// <param name="source">The source to restore collisions with.</param>
+        protected virtual void OnSourceRemoved(GameObject source)
+        {
+            ToggleCollisions(source, Sources, Targets, false);
+        }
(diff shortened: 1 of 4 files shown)


thestonefox:
@bddckr  Dunno if you'll get a "re-review" request on this, but consider this it :)

Runtime/Tracking/Collision/CollisionIgnorer.cs
```diff
@@ -0,0 +1,213 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Sets the collision state between the source and targets.
+        /// </summary>
+        /// <param name="source">The source to set the collision state on.</param>
+        /// <param name="targets">A collection of targets to set the collision state on.</param>
+        /// <param name="state">Whether to ignore collisions or not.</param>
+        protected virtual void ToggleCollisions(GameObject source, GameObjectObservableList sources, GameObjectObservableList targets, bool state)
```
bddckr: Missing documentation for the `sources` parameter.

Runtime/Tracking/Collision/CollisionIgnorer.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            if (source == null || (!state && sources.Contains(source)))
+            {
+                return;
+            }
+
+            List<Collider> sourceColliders = new List<Collider>();
+            List<Collider> targetColliders = new List<Collider>();
+
+            sourceColliders.AddRange(source.GetComponentsInChildren<Collider>());
```
bddckr: There's an overload of the `GetComponentsInChildren` method that *takes* a list. Use that for both these calls together with two lists you reuse by making it a `protected` field instead. Note that you do **not** have to call `Clear` on those lists - `GetComponentsInChildren` does that for you already.

This means we reduce the amount of GC allocations since the list will not change in capacity (not size) all the time.

Runtime/Tracking/Collision/CollisionIgnorer.cs
```diff
@@ -0,0 +1,213 @@
...
+
+            foreach (GameObject target in targets.SubscribableElements)
+            {
+                if (target != null)
```
bddckr: Didn't we say we want `null` values to be treated as errors and can thus ignore it here as the null ref exception being thrown without the if is enough for the user to know?

Runtime/Tracking/Collision/CollisionIgnorer.cs
```diff
@@ -0,0 +1,213 @@
...
+        protected virtual void AddTarget(GameObject target)
+        {
+            ToggleCollisions(target, Targets, Sources, true);
+        }
+
+        /// <summary>
+        /// Removes an existing target from having collisions ignored with all of the <see cref="sources"/> colliders.
+        /// </summary>
+        /// <param name="target">The target to restore collisions on.</param>
+        public virtual void RemoveTarget(GameObject target)
```
bddckr:
```suggestion
        protected virtual void RemoveTarget(GameObject target)
```

Runtime/Tracking/Collision/CollisionIgnorer.cs
```diff
@@ -0,0 +1,213 @@
...
+            UnregisterTargetListeners();
+            ToggleCollisions(false);
+        }
+
+        /// <summary>
+        /// Registers the listeners for elements that are added or removed from <see cref="Sources"/>.
+        /// </summary>
+        protected virtual void RegisterSourceListeners()
+        {
+            Sources.ElementAdded.AddListener(AddSource);
```
bddckr:
```suggestion
            Sources.ElementAdded.AddListener(OnSourceAdded);
```
I don't like the verb name for that method. It's a callback, not doing an add operation.

**Same for the Remove callback methods for both the Source and Target.** The XML documentation comments also need to change on those methods to explain they're callbacks and actually don't "do that operation".

Tests/Editor/Tracking/Collision/CollisionIgnorerTest.cs
```diff
@@ -0,0 +1,193 @@
...
+        [TearDown]
+        public void TearDown()
+        {
+            Object.DestroyImmediate(containingObject);
+        }
+
+        [UnityTest]
+        public IEnumerator CollisionsIgnored()
+        {
+            WaitForFixedUpdate collisionOccurredYieldInstruction = new WaitForFixedUpdate();
```
bddckr: Given this naming I first thought you created a custom `YieldInstruction` subclass that waits until the collision flag is set!

Perhaps we rename this?
```suggestion
            WaitForFixedUpdate waitForFixedUpdate = new WaitForFixedUpdate();
```

Runtime/Tracking/Collision/CollisionIgnorer.cs
thestonefox: Not in this case, because it's called OnDisable and the target can be destroyed when closing and if this runs still and the target is null (because it's been destroyed) then it will throw an error

Tests/Editor/Tracking/Collision/CollisionIgnorerTest.cs
thestonefox: the reason it was named the way it was, because the WaitForFixedUpdate instruction kept changing whilst i found a good actual instruction that was consistent. 

But fixed update is consistent so i'm happy to rename the variable

Runtime/Tracking/Collision/CollisionIgnorer.cs
bddckr: 👍

Runtime/Tracking/Collision/CollisionIgnorer.cs
thestonefox: I wouldn't say these are simply callbacks. They are actually carrying out a function. You could call one of these methods directly (for instance in an inherited class) that passes a GameObject into that method.

On that, i'd say the AddSource is still probably correct naming.

Runtime/Tracking/Collision/CollisionIgnorer.cs
thestonefox: code please :) not sure what you mean :p

Runtime/Tracking/Collision/CollisionIgnorer.cs
bddckr: It's not adding a source, though. Perhaps `IgnoreSource` tells what it does better, but I'm not sure about the equivalent for `Remove` then. `UnignoreSource` seems strange 😄

Runtime/Tracking/Collision/CollisionIgnorer.cs
bddckr: https://github.com/ExtendRealityLtd/Zinnia.Unity/blob/0649c35f7d920424d4caa59abc10df3f9b2b3092/Runtime/Tracking/Collision/CollisionNotifier.cs#L174-L177
https://github.com/ExtendRealityLtd/Zinnia.Unity/blob/0649c35f7d920424d4caa59abc10df3f9b2b3092/Runtime/Tracking/Collision/CollisionNotifier.cs#L207

Runtime/Tracking/Collision/CollisionIgnorer.cs
thestonefox: `AddSource` is actually ignoring the given gameobejct against the Targets list.

I guess anyone that wants to replicate the result of the method can just directly call the `ToggleCollisions` method and these can be purely used for the listeners so rename them to `OnSourceAdded` etc.

Runtime/Tracking/Collision/CollisionIgnorer.cs
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        /// <summary>
+        /// Sets the collision state between the source and targets.
+        /// </summary>
+        /// <param name="source">The source to set the collision state on.</param>
+        /// <param name="sources">A collection of sources to check if the given <see cref="source"/> belongs to.</param>
+        /// <param name="targets">A collection of targets to set the collision state on.</param>
+        /// <param name="state">Whether to ignore collisions or not.</param>
+        protected virtual void ToggleCollisions(GameObject source, GameObjectObservableList sources, GameObjectObservableList targets, bool state)
```
bddckr: `targetColliders` will be cleared and filled again each time you call it, sorry I didn't explain how that API works (and it's undocumented as far as I can see 🙄). To solve it let's do this:
```c#
protected virtual void ToggleCollisions(GameObject source, GameObjectObservableList sources, GameObjectObservableList targets, bool state)
{
    if (source == null || (!state && sources.Contains(source)))
    {
        return;
    }

    source.GetComponentsInChildren(sourceColliders);

    foreach (GameObject target in targets.SubscribableElements)
    {
        if (target == null)
        {
            continue;
        }

        target.GetComponentsInChildren(targetColliders);

        foreach (Collider sourceCollider in sourceColliders)
        {
            foreach (Collider targetCollider in targetColliders)
            {
                Physics.IgnoreCollision(sourceCollider, targetCollider, state);
            }
        }
    }
}
```

PR #304, opened 2019-03-04
feat(Visual): update all fields to serialized properties

thestonefox:
All usages of a field have now been updated to use a serialized
property to allow setting of data via UnityEvents.

New helper methods have been implemented to allow changes to
properties via code or the inspector to react to the changes and
update any relevant components as required.

The CameraColorOverlay component now also uses a working material to
apply the color overlay to instead of working directly on the injected
material. This means the injected material is used as a blueprint but
is never changed at runtime.

This new working material is automatically updated anytime the
OverlayMaterial property is changed.

### DIFF
--- a/Runtime/Visual/CameraColorOverlay.cs
+++ b/Runtime/Visual/CameraColorOverlay.cs
@@ -5,6 +5,9 @@
     using System;
     using System.Collections;
     using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
+    using Malimbe.BehaviourStateRequirementMethod;
+    using Malimbe.MemberChangeMethod;
     using Zinnia.Extension;
     using Zinnia.Data.Type;
 
@@ -22,17 +25,18 @@
             /// <summary>
             /// The <see cref="Color"/> being applied to the camera overlay.
             /// </summary>
-            [DocumentedByXml]
-            public Color color;
+            [Serialized]
+            [field: DocumentedByXml]
+            public Color Color { get; set; }
 
             public EventData Set(EventData source)
             {
-                return Set(source.color);
+                return Set(source.Color);
             }
 
             public EventData Set(Color color)
             {
-                this.color = color;
+                Color = color;
                 return this;
             }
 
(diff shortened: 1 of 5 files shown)


Runtime/Visual/CameraColorOverlay.cs
```diff
@@ -22,17 +25,17 @@ public class EventData
             /// <summary>
             /// The <see cref="Color"/> being applied to the camera overlay.
             /// </summary>
-            [DocumentedByXml]
-            public Color color;
+            [Serialized]
```
bddckr:
```suggestion
            [Serialized]
            [field: DocumentedByXml]
```

Runtime/Visual/CameraColorOverlay.cs
```diff
@@ -252,5 +269,23 @@ protected virtual void PostRender(Camera cam)
                 GL.PopMatrix();
             }
         }
+
+        /// <summary>
+        /// Copies the <see cref="OverlayMaterial"/> data to the <see cref="workingMaterial"/> data.
+        /// </summary>
+        protected virtual void CopyMaterialOverlayToWorking()
+        {
+            workingMaterial = new Material(OverlayMaterial);
```
bddckr: This needs to destroy the previous `workingMaterial` before creating a new instance (if it's not `null` that is).

Runtime/Visual/CameraColorOverlay.cs
```diff
@@ -252,5 +269,23 @@ protected virtual void PostRender(Camera cam)
...
+        }
+
+        /// <summary>
+        /// Called after the <see cref="OverlayMaterial"/> property has been changed.
```
bddckr:
```suggestion
        /// Called after <see cref="OverlayMaterial"/> has been changed.
```

Runtime/Visual/CameraColorOverlay.cs
```diff
@@ -252,5 +269,23 @@ protected virtual void PostRender(Camera cam)
...
+        /// </summary>
+        [CalledAfterChangeOf(nameof(OverlayMaterial))]
+        [RequiresBehaviourState]
```
bddckr: Code convention: Merge attributes.

Runtime/Visual/PointsRenderer.cs
```diff
@@ -20,38 +22,53 @@ public class PointsData
             /// <summary>
             /// Represents the start, i.e. the first rendered point.
             /// </summary>
-            public GameObject start;
+            [Serialized]
```
bddckr:
```suggestion
            [Serialized]
            [field: DocumentedByXml]
```

Runtime/Visual/PointsRenderer.cs
```diff
@@ -20,38 +22,53 @@ public class PointsData
             /// <summary>
             /// Represents the start, i.e. the first rendered point.
             /// </summary>
-            public GameObject start;
+            [Serialized]
+            public GameObject PointStart { get; set; }
```
bddckr: Changing all of these new names to have `Point` at the end reads better to me. `StartPoint`, `RepeatedSegmentPoint`, `EndPoint`.

Runtime/Visual/PointsRenderer.cs
```diff
@@ -20,38 +22,53 @@ public class PointsData
...
             /// </summary>
-            public GameObject repeatedSegment;
+            [Serialized]
+            public GameObject PointRepeatedSegment { get; set; }
             /// <summary>
             /// Represents the end, i.e. the last rendered point.
             /// </summary>
-            public GameObject end;
+            [Serialized]
+            public GameObject PointEnd{ get; set; }
```
bddckr:
```suggestion
            public GameObject PointEnd { get; set; }
```

Runtime/Visual/PointsRenderer.cs
```diff
@@ -20,38 +22,53 @@ public class PointsData
...
             /// </summary>
-            public GameObject start;
+            [Serialized]
+            public GameObject PointStart { get; set; }
             /// <summary>
-            /// Represents the segments between <see cref="start"/> and <see cref="end"/>. This will get cloned to create all the segments.
+            /// Represents the segments between <see cref="Start"/> and <see cref="End"/>. This will get cloned to create all the segments.
             /// </summary>
-            public GameObject repeatedSegment;
+            [Serialized]
```
bddckr:
```suggestion
            [Serialized]
            [field: DocumentedByXml]
```

Runtime/Visual/PointsRenderer.cs
```diff
@@ -20,38 +22,53 @@ public class PointsData
...
+            public GameObject PointRepeatedSegment { get; set; }
             /// <summary>
             /// Represents the end, i.e. the last rendered point.
             /// </summary>
-            public GameObject end;
+            [Serialized]
```
bddckr:
```suggestion
            [Serialized]
            [field: DocumentedByXml]
```

Runtime/Visual/PointsRenderer.cs
```diff
@@ -20,38 +22,53 @@ public class PointsData
...
+            public GameObject PointEnd{ get; set; }
             /// <summary>
             /// The points along the the most recent cast.
             /// </summary>
-            public IReadOnlyList<Vector3> points;
+            [Serialized]
```
bddckr:
```suggestion
            [Serialized]
            [field: DocumentedByXml]
```

Runtime/Visual/CameraColorOverlay.cs
thestonefox: Why do we need to destroy the previous one? Won't this simply overwrite the new one and the old one will be garbage collected?

bddckr: No, because `Material` is a Unity "resource" and as such needs to be destroyed on their native engine side.

bddckr: (It shows that by inheriting from `UnityEngine.Object` - all of those need to be `Destroy`ed.)

PR #309, opened 2019-03-05
Fix/property drawers

thestonefox:


Editor/Data/Attribute/RestrictedAttributeDrawer.cs
```diff
@@ -0,0 +1,63 @@
...
+    using UnityEngine;
+    using UnityEditor;
+
+    [CustomPropertyDrawer(typeof(RestrictedAttribute))]
+    public class RestrictedAttributeDrawer : PropertyDrawer
+    {
+        public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
+        {
+            EditorGUI.BeginProperty(position, label, property);
+            RestrictedAttribute attrib = attribute as RestrictedAttribute;
```
bddckr:
```suggestion
            RestrictedAttribute attrib = (RestrictedAttribute)attribute;
```
I don't like the short variable name but can't come up with something else than `restrictedAttribute` which is a bit long.

Editor/Data/Attribute/RestrictedAttributeDrawer.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            bool playingAndActiveAndEnabled = Application.isPlaying && property.serializedObject.targetObject is Behaviour behaviour && behaviour.isActiveAndEnabled;
```
bddckr:
```suggestion
            bool isPlayingAndActiveAndEnabled = Application.isPlaying && property.serializedObject.targetObject is Behaviour behaviour && behaviour.isActiveAndEnabled;
```
Regular C# naming convention for boolean types.

Editor/Data/Attribute/RestrictedAttributeDrawer.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            bool originalGUIEnabledState = GUI.enabled;
+            if (makeReadOnly)
+            {
+                GUI.enabled = false;
+            }
+
+            Color controlColor = Color.grey;
+
+            Color currentGUIColor = GUI.color;
```
bddckr:
```suggestion
            Color currentGuiColor = GUI.color;
```

Editor/Data/Attribute/RestrictedAttributeDrawer.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            bool muteProperty = (attrib.restrictions & RestrictedAttribute.Restrictions.Muted) != 0;
+
+            bool originalGUIEnabledState = GUI.enabled;
+            if (makeReadOnly)
+            {
+                GUI.enabled = false;
+            }
+
+            Color controlColor = Color.grey;
```
bddckr:
```suggestion
            const Color controlColor = Color.grey;
```
I think that works.

Editor/Data/Attribute/RestrictedAttributeDrawer.cs
bddckr: I suggest to rename all of these `current*` variables to `original*` like the other one above.

Editor/Data/Attribute/RestrictedAttributeDrawer.cs
```diff
@@ -0,0 +1,63 @@
...
+                GUI.color = currentGUIColor;
+                EditorStyles.label.normal.textColor = currentNormalTextColor;
+                EditorStyles.label.focused.textColor = currentFocusedTextColor;
+                EditorStyles.label.fontStyle = currentFontStyle;
+            }
+
+            if (makeReadOnly)
+            {
+                GUI.enabled = originalGUIEnabledState;
+            }
```
bddckr:
```suggestion
            }

```
I feel like we always split control statements with the following expressions with some space like this, but your call!

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Zinnia.Data.Attribute
+{
+    using UnityEngine;
+    using System;
+
+    /// <summary>
+    /// Defines the <c>[Restricted]</c> attribute.
```
bddckr:
```suggestion
    /// Draws the member in a specified restricted way.
```
We should actually fix this on all attributes...

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    public class RestrictedAttribute : PropertyAttribute
+    {
+        [Flags]
+        public enum Restrictions
```
bddckr: Anything in this class needs docs: Enum, enum's variants, field, constructor. The constructor's summary should be a copy of the type's summary.

Editor/Data/Attribute/RestrictedAttributeDrawer.cs
thestonefox: Color can't be const

bddckr: 👍

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class RestrictedAttribute : PropertyAttribute
+    {
+        /// <summary>
+        /// The restriction options that can be applied.
+        /// </summary>
+        [Flags]
+        public enum Restrictions
+        {
+            /// <summary>
+            /// The property is always set to ReadOnly
```
bddckr:
```suggestion
            /// The property is always read-only.
```

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace Zinnia.Data.Attribute
+{
+    using UnityEngine;
+    using System;
+
+    /// <summary>
+    /// Draws the member in a specified restricted way.
```
bddckr:
```suggestion
    /// Draws the property in a specified restricted way.
```
You used that in the docs you added and I think that's what Unity uses, too.

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        [Flags]
+        public enum Restrictions
+        {
+            /// <summary>
+            /// The property is always set to ReadOnly
+            /// </summary>
+            ReadOnlyAlways = 1 << 0,
+            /// <summary>
+            /// The property is only set to ReadOnly when the Application is playing and the component is enabled.
```
bddckr:
```suggestion
            /// The property is read-only when the application is playing and the component is enabled.
```

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+            /// </summary>
+            ReadOnlyWhenPlayingAndEnabled = 1 << 1,
+            /// <summary>
+            /// The property style is de-emphasized.
```
bddckr:
```suggestion
            /// The property is de-emphasized.
```

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+            /// </summary>
+            Muted = 1 << 2
+        }
+
+        /// <summary>
+        /// The restriction options that will be applied.
```
bddckr:
```suggestion
        /// The restriction options to apply.
```

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        public readonly Restrictions restrictions;
+
+        /// <summary>
+        /// Draws the member in a specified restricted way.
```
bddckr:
```suggestion
        /// Draws the property in a specified restricted way.
```

PR #310, opened 2019-03-05
feat(Utility): update all fields to be serialized properties

thestonefox:
All usages of fields have now been switched to properties that
are serialized with Malimbe to create an appropriate backing field
to display in the Unity Inspector.

New helper methods have been implemented to allow changes to
properties via code or the inspector to react to the changes and
update any relevant components as required.

Also, now the tests run in play mode, the CountdownTimer can be
tested so a new test has been added to unit test the functions
of the CountdownTimer component.

### DIFF
--- a/Runtime/Utility/CountdownTimer.cs
+++ b/Runtime/Utility/CountdownTimer.cs
@@ -1,9 +1,10 @@
 ﻿namespace Zinnia.Utility
 {
-    using Malimbe.BehaviourStateRequirementMethod;
-    using Malimbe.XmlDocumentationAttribute;
     using UnityEngine;
     using UnityEngine.Events;
+    using Malimbe.BehaviourStateRequirementMethod;
+    using Malimbe.PropertySerializationAttribute;
+    using Malimbe.XmlDocumentationAttribute;
 
     /// <summary>
     /// Counts down from a given start time until zero and emits appropriate events throughout the process.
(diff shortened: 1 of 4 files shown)


thestonefox:
This commit contains the fix for the InterfaceContainer too so is required for any tests to pass that use the InterfaceContainer

Runtime/Utility/CountdownTimer.cs
```diff
@@ -67,9 +69,12 @@ public virtual void Begin()
...
         public virtual void Cancel()
         {
-            IsRunning = false;
             CancelInvoke(nameof(Complete));
-            Cancelled?.Invoke();
+            if (IsRunning)
+            {
+                Cancelled?.Invoke();
+            }
+            IsRunning = false;
```
bddckr: Move this into the if to save an assignment that doesn't do anything as it's already false.

Runtime/Utility/InterfaceContainer.cs
```diff
@@ -35,6 +30,9 @@ protected virtual void OnFieldChange(Object previousValue, ref Object newValue)
     /// <typeparam name="TInterface">The interface type to contain.</typeparam>
     public abstract class InterfaceContainer<TInterface> : InterfaceContainer, ISerializationCallbackReceiver
     {
+        /// <summary>
+        /// The interface container.
```
bddckr:
```suggestion
        /// The contained interface.
```

Tests/Editor/Utility/CountdownTimerTest.cs
```diff
@@ -0,0 +1,322 @@
...
+        public void SetUp()
+        {
+            containingObject = new GameObject();
+            subject = containingObject.AddComponent<CountdownTimer>();
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            Object.Destroy(subject);
```
bddckr:
```suggestion
```
Destroying the game object also deletes any attached component.

thestonefox:
Updated and force pushed

PR #311, opened 2019-03-06
feat(Collection): add extra functionality to ObservableList

thestonefox:
The ObservableList component now has additional functionality like
being able to check if the collection contains a given item and
set an element value at a specific index.

The fields have now been converted to properties and a new CurrentIndex
field has been added that allows methods that require an index to
still function if called via the limitations of a UnityEvent (which
only allows one parameter type).

### DIFF
--- a/Runtime/Data/Collection/ObservableList.cs
+++ b/Runtime/Data/Collection/ObservableList.cs
@@ -5,6 +5,8 @@
     using System.Collections.Generic;
     using Malimbe.BehaviourStateRequirementMethod;
     using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
+    using Zinnia.Extension;
 
     /// <summary>
     /// Allows observing changes to a <see cref="List{T}"/>.
@@ -16,9 +18,27 @@
         /// <summary>
         /// The collection to observe changes of.
         /// </summary>
-        [SerializeField, DocumentedByXml]
-        protected List<TElement> elements = new List<TElement>();
+        [Serialized]
+        [field: DocumentedByXml]
+        protected List<TElement> Elements { get; set; } = new List<TElement>();
 
+        /// <summary>
+        /// The index to use in methods specifically specifying to use it. In case this index is out of bounds for the collection it will wrap around.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public int CurrentIndex { get; set; }
+
+        /// <summary>
+        /// Emitted when the searched element is found.
+        /// </summary>
+        [DocumentedByXml]
+        public TEvent Found = new TEvent();
+        /// <summary>
+        /// Emitted when the searched element is not found.
+        /// </summary>
+        [DocumentedByXml]
+        public TEvent NotFound = new TEvent();
         /// <summary>
         /// Emitted when the first element is added to the collection.
         /// </summary>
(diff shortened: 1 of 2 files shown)


Runtime/Data/Collection/ObservableList.cs
```diff
@@ -16,9 +18,27 @@ public abstract class ObservableList<TElement, TEvent> : MonoBehaviour where TEv
...
         /// The collection to observe changes of.
         /// </summary>
-        [SerializeField, DocumentedByXml]
-        protected List<TElement> elements = new List<TElement>();
+        [Serialized]
+        [field: DocumentedByXml]
+        protected List<TElement> Elements { get; set; } = new List<TElement>();
 
+        /// <summary>
+        /// The pointer for the current index being specified in the collection.
```
bddckr:
```suggestion
        /// The index to use in methods specifically specifying to use it. In case this index is out of bounds for the collection it will wrap around.
```

Runtime/Data/Collection/ObservableList.cs
```diff
@@ -16,9 +18,27 @@ public abstract class ObservableList<TElement, TEvent> : MonoBehaviour where TEv
...
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public int CurrentIndex { get; set; }
+
+        /// <summary>
+        /// Emitted when the searched element is found.
+        /// </summary>
+        [DocumentedByXml]
+        public TEvent Found = new TEvent();
```
bddckr: No properties? Can't remember our decision.

Runtime/Data/Collection/ObservableList.cs
```diff
@@ -43,38 +63,120 @@ public abstract class ObservableList<TElement, TEvent> : MonoBehaviour where TEv
...
 
         /// <summary>
-        /// Adds an element to the end of the collection.
+        /// Adds an element to the given index of the collection.
         /// </summary>
+        /// <remarks>
+        /// Allows the use of a wrapped and clamped index to prevent indices being out of bounds and doing negative queries such as `-1` sets the last element.
+        /// </remarks>
         /// <param name="element">The element to add.</param>
+        /// <param name="index">The index to add at.</param>
```
bddckr:
```suggestion
        /// <param name="index">The index to add at. In case this index is out of bounds for the collection it will wrap around.</param>
```

Runtime/Data/Collection/ObservableList.cs
```diff
@@ -43,38 +63,120 @@ public abstract class ObservableList<TElement, TEvent> : MonoBehaviour where TEv
...
+        }
+
+        /// <summary>
+        /// Sets the given element at the given index.
+        /// </summary>
+        /// <remarks>
+        /// Allows the use of a wrapped and clamped index to prevent indices being out of bounds and doing negative queries such as `-1` sets the last element.
+        /// </remarks>
+        /// <param name="element">The element to set.</param>
+        /// <param name="index">The index in the collection to set at.</param>
```
bddckr:
```suggestion
        /// <param name="index">The index in the collection to set at. In case this index is out of bounds for the collection it will wrap around.</param>
```

Runtime/Data/Collection/ObservableList.cs
```diff
@@ -85,16 +187,45 @@ public virtual void RemoveFirst(TElement element)
...
+            RemoveAt(index);
+        }
+
+        /// <summary>
+        /// Removes an element at the given index from the collection.
+        /// </summary>
+        /// <remarks>
+        /// Allows the use of a wrapped and clamped index to prevent indices being out of bounds and doing negative queries such as `-1` sets the last element.
+        /// </remarks>
+        /// <param name="index">The index to remove at.</param>
```
bddckr:
```suggestion
        /// <param name="index">The index to remove at. In case this index is out of bounds for the collection it will wrap around.</param>
```

Runtime/Data/Collection/ObservableList.cs
thestonefox: There's no benefit making the UnityEvent a property is there? you can't set anything with them and the unity standard is to make them fields.

thestonefox:
Updated and force pushed

Runtime/Data/Collection/ObservableList.cs
bddckr: 👍

PR #315, opened 2019-03-07
feat(Rule): update fields to properties and lists to observable

thestonefox:

### DIFF
--- a/Runtime/Rule/ListContainsRule.cs
+++ b/Runtime/Rule/ListContainsRule.cs
@@ -1,9 +1,10 @@
 ﻿namespace Zinnia.Rule
 {
     using UnityEngine;
-    using System.Collections.Generic;
     using Malimbe.XmlDocumentationAttribute;
     using Malimbe.BehaviourStateRequirementMethod;
+    using Malimbe.PropertySerializationAttribute;
+    using Zinnia.Data.Collection;
 
     /// <summary>
     /// Determines whether an object is part of a list.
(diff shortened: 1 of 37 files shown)


Runtime/Rule/ListContainsRule.cs
```diff
@@ -13,15 +14,16 @@ public class ListContainsRule : MonoBehaviour, IRule
...
+        [field: DocumentedByXml]
+        public UnityObjectObservableList Objects { get; set; }
 
         /// <inheritdoc />
         [RequiresBehaviourState]
         public virtual bool Accepts(object target)
         {
             Object targetObject = target as Object;
-            return targetObject != null && objects.Contains(targetObject);
+            return targetObject != null && Objects != null && Objects.Contains(targetObject);
```
bddckr:
```suggestion
            return targetObject != null && Objects.Contains(targetObject);
```

Runtime/Rule/ListContainsRule.cs
```diff
@@ -13,15 +14,16 @@ public class ListContainsRule : MonoBehaviour, IRule
...
-        [DocumentedByXml]
-        public List<Object> objects = new List<Object>();
+        [Serialized]
+        [field: DocumentedByXml]
+        public UnityObjectObservableList Objects { get; set; }
 
         /// <inheritdoc />
         [RequiresBehaviourState]
         public virtual bool Accepts(object target)
         {
```
bddckr:
```suggestion
         {
             if (Objects == null)
             {
                 return false;
             }
```
This saves the cast and null-check for `targetObject` below.

PR #319, opened 2019-03-09
feat(Haptics):  update fields to properties and lists to observable

thestonefox:

### DIFF
--- a/Runtime/Haptics/AudioClipHapticPulser.cs
+++ b/Runtime/Haptics/AudioClipHapticPulser.cs
@@ -3,34 +3,38 @@
     using UnityEngine;
     using System.Collections;
     using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
 
     /// <summary>
-    /// Creates a haptic pattern based on the waveform of an <see cref="UnityEngine.AudioClip"/> and utilizes a <see cref="HapticPulser"/> to create the effect.
+    /// Creates a haptic pattern based on the waveform of an <see cref="UnityEngine.AudioClip"/> and utilizes a <see cref="Haptics.HapticPulser"/> to create the effect.
     /// </summary>
     public class AudioClipHapticPulser : HapticProcess
     {
         /// <summary>
         /// The pulse process to utilize.
         /// </summary>
-        [DocumentedByXml]
-        public HapticPulser hapticPulser;
+        [Serialized]
+        [field: DocumentedByXml]
+        public HapticPulser HapticPulser { get; set; }
         /// <summary>
         /// The waveform to represent the haptic pattern.
         /// </summary>
-        [DocumentedByXml]
-        public AudioClip audioClip;
+        [Serialized]
+        [field: DocumentedByXml]
+        public AudioClip AudioClip { get; set; }
         /// <summary>
         /// Multiplies the current audio peak to affect the wave intensity.
         /// </summary>
-        [DocumentedByXml]
-        public float intensityMultiplier = 1f;
+        [Serialized]
+        [field: DocumentedByXml]
+        public float IntensityMultiplier { get; set; } = 1f;
 
         /// <summary>
         /// A reference to the started routine.
         /// </summary>
         protected Coroutine hapticRoutine;
         /// <summary>
-        /// The original intensity of the provided <see cref="HapticPulser"/> to reset back to after the process is complete.
+        /// The original intensity of <see cref="HapticPulser"/> to reset back to after the process is complete.
         /// </summary>
         protected float cachedIntensity;
         /// <summary>
@@ -41,13 +45,13 @@
         /// <inheritdoc />
         public override bool IsActive()
         {
-            return (base.IsActive() && hapticPulser != null && audioClip != null && hapticPulser.IsActive());
+            return base.IsActive() && HapticPulser != null && AudioClip != null && HapticPulser.IsActive();
         }
 
         /// <inheritdoc />
         protected override void DoBegin()
         {
-            cachedIntensity = hapticPulser.Intensity;
+            cachedIntensity = HapticPulser.Intensity;
             hapticRoutine = StartCoroutine(HapticProcessRoutine());
         }
 
@@ -61,7 +65,7 @@
 
             StopCoroutine(hapticRoutine);
             hapticRoutine = null;
-            hapticPulser.Cancel();
+            HapticPulser.Cancel();
             ResetIntensity();
         }
 
@@ -70,11 +74,11 @@
         /// </summary>
         protected virtual void ResetIntensity()
         {
-            hapticPulser.Intensity = cachedIntensity;
+            HapticPulser.Intensity = cachedIntensity;
         }
 
         /// <summary>
-        /// Enumerates through the given <see cref="AudioClip"/> and pulses for each amplitude of the wave.
+        /// Enumerates through <see cref="AudioClip"/> and pulses for each amplitude of the wave.
         /// </summary>
         /// <returns>An Enumerator to manage the running of the Coroutine.</returns>
         protected virtual IEnumerator HapticProcessRoutine()
@@ -82,21 +86,21 @@
             float[] audioData = new float[BufferSize];
             int sampleOffset = -BufferSize;
             float startTime = Time.time;
-            float length = audioClip.length;
+            float length = AudioClip.length;
             float endTime = startTime + length;
-            float sampleRate = audioClip.samples;
+            float sampleRate = AudioClip.samples;
             while (Time.time <= endTime)
             {
                 float lerpVal = (Time.time - startTime) / length;
                 int sampleIndex = (int)(sampleRate * lerpVal);
                 if (sampleIndex >= sampleOffset + BufferSize)
                 {
-                    audioClip.GetData(audioData, sampleIndex);
+                    AudioClip.GetData(audioData, sampleIndex);
                     sampleOffset = sampleIndex;
                 }
                 float currentSample = Mathf.Abs(audioData[sampleIndex - sampleOffset]);
-                hapticPulser.Intensity = currentSample * intensityMultiplier;
-                hapticPulser.Begin();
+                HapticPulser.Intensity = currentSample * IntensityMultiplier;
+                HapticPulser.Begin();
                 yield return null;
             }
             ResetIntensity();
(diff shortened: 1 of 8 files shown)


Runtime/Haptics/AudioClipHapticPulser.cs
```diff
@@ -2,35 +2,40 @@
...
+        [field: DocumentedByXml]
+        public float IntensityMultiplier { get; set; } = 1f;
 
         /// <summary>
         /// A reference to the started routine.
         /// </summary>
         protected Coroutine hapticRoutine;
         /// <summary>
-        /// The original intensity of the provided <see cref="HapticPulser"/> to reset back to after the process is complete.
+        /// The original intensity of the provided <see cref="Haptics.HapticPulser"/> to reset back to after the process is complete.
```
bddckr:
```suggestion
        /// The original intensity of <see cref="HapticPulser"/> to reset back to after the process is complete.
```

Runtime/Haptics/AudioClipHapticPulser.cs
```diff
@@ -2,35 +2,40 @@
...
+    /// Creates a haptic pattern based on the waveform of an <see cref="UnityEngine.AudioClip"/> and utilizes a <see cref="Haptics.HapticPulser"/> to create the effect.
     /// </summary>
     public class AudioClipHapticPulser : HapticProcess
     {
         /// <summary>
         /// The pulse process to utilize.
         /// </summary>
-        [DocumentedByXml]
-        public HapticPulser hapticPulser;
+        [Serialized, Cleared]
```
bddckr:
```suggestion
        [Serialized]
```

Runtime/Haptics/AudioClipHapticPulser.cs
```diff
@@ -2,35 +2,40 @@
...
+        [field: DocumentedByXml]
+        public HapticPulser HapticPulser { get; set; }
         /// <summary>
         /// The waveform to represent the haptic pattern.
         /// </summary>
-        [DocumentedByXml]
-        public AudioClip audioClip;
+        [Serialized, Cleared]
```
bddckr:
```suggestion
        [Serialized]
```

Runtime/Haptics/AudioClipHapticPulser.cs
```diff
@@ -70,33 +75,33 @@ protected override void DoCancel()
         /// </summary>
         protected virtual void ResetIntensity()
         {
-            hapticPulser.Intensity = cachedIntensity;
+            HapticPulser.Intensity = cachedIntensity;
         }
 
         /// <summary>
-        /// Enumerates through the given <see cref="AudioClip"/> and pulses for each amplitude of the wave.
+        /// Enumerates through the given <see cref="UnityEngine.AudioClip"/> and pulses for each amplitude of the wave.
```
bddckr:
```suggestion
        /// Enumerates through <see cref="AudioClip"/> and pulses for each amplitude of the wave.
```

Runtime/Haptics/HapticPulser.cs
```diff
@@ -14,19 +13,17 @@ public abstract class HapticPulser : HapticProcess
         /// <summary>
         /// The intensity of the haptic rumble.
         /// </summary>
-        [Serialized, /*Validated*/]
-        [field: Range(0f, 1f), DocumentedByXml]
+        [Serialized,]
```
bddckr:
```suggestion
        [Serialized]
```

Runtime/Haptics/HapticPulser.cs
```diff
@@ -1,10 +1,9 @@
 ﻿namespace Zinnia.Haptics
 {
-    using Malimbe.PropertySerializationAttribute;
-    /*using Malimbe.PropertySetterMethod;*/
-    /*using Malimbe.PropertyValidationMethod;*/
-    using Malimbe.XmlDocumentationAttribute;
     using UnityEngine;
+    using Malimbe.XmlDocumentationAttribute;
```
bddckr: The order of these three imports isn't correct. Let's order in alphabetical order.

Runtime/Haptics/HapticPulser.cs
```diff
@@ -14,19 +13,17 @@ public abstract class HapticPulser : HapticProcess
...
         /// </summary>
-        /// <param name="previousValue">The previous value.</param>
-        /// <param name="newValue">The new value.</param>
-        /*[CalledBySetter(nameof(Intensity))]*/
-        protected virtual void OnIntensityChange(float previousValue, ref float newValue)
+        [CalledAfterChangeOf(nameof(Intensity))]
+        protected virtual void OnAfterIntensityChange()
         {
-            newValue = Mathf.Clamp01(newValue);
+            Intensity = Mathf.Clamp01(Intensity);
```
bddckr: This loops infinitely - check if the desired result `Mathf.Clamp01(Intensity)` isn't already the value of `Intensity`.

Runtime/Haptics/TimedHapticProcess.cs
```diff
@@ -2,32 +2,33 @@
 {
     using UnityEngine;
     using System.Collections;
-    using Malimbe.PropertySerializationAttribute;
-    /*using Malimbe.PropertyValidationMethod;*/
+    using Malimbe.MemberClearanceMethod;
```
bddckr: Alphabetical order.

Runtime/Haptics/AudioClipHapticPulser.cs
```diff
@@ -41,13 +46,13 @@ public class AudioClipHapticPulser : HapticProcess
         /// <inheritdoc />
         public override bool IsActive()
         {
-            return (base.IsActive() && hapticPulser != null && audioClip != null && hapticPulser.IsActive());
+            return (base.IsActive() && HapticPulser != null && AudioClip != null && HapticPulser.IsActive());
```
bddckr:
```suggestion
            return base.IsActive() && HapticPulser != null && AudioClip != null && HapticPulser.IsActive();
```
Might as well while we're touching this. 😄

Runtime/Haptics/TimedHapticProcess.cs
```diff
@@ -62,11 +63,11 @@ protected override void DoCancel()
 
             StopCoroutine(hapticRoutine);
             hapticRoutine = null;
-            hapticProcess.Cancel();
+            HapticProcess.Cancel();
         }
 
         /// <summary>
-        /// Enumerates for the specified duration calling the given <see cref="hapticProcess"/> with a specified interval delay between each call.
+        /// Enumerates for the specified duration calling the given <see cref="HapticProcess"/> with a specified interval delay between each call.
```
bddckr:
```suggestion
        /// Enumerates for the specified duration calling <see cref="HapticProcess"/> with a specified interval delay between each call.
```

Runtime/Haptics/XRNodeHapticPulser.cs
```diff
@@ -1,7 +1,8 @@
 ﻿namespace Zinnia.Haptics
 {
-    using Malimbe.XmlDocumentationAttribute;
     using UnityEngine.XR;
+    using Malimbe.XmlDocumentationAttribute;
```
bddckr: Alphabetical order.

Runtime/Haptics/AudioClipHapticPulser.cs
```diff
@@ -2,35 +2,40 @@
 {
     using UnityEngine;
     using System.Collections;
+    using Malimbe.MemberClearanceMethod;
```
bddckr: Alphabetical order.

Runtime/Haptics/HapticProcessor.cs
```diff
@@ -1,7 +1,8 @@
 ﻿namespace Zinnia.Haptics
 {
-    using System.Collections.Generic;
     using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
```
bddckr: Alphabetical order.

Runtime/Haptics/AudioClipHapticPulser.cs
thestonefox: I've been putting them in order of string length (so it looks neat and pleasing to the eye :) )

bddckr: Heh, sadly I can't tell my auto-formatting of the IDE **that**.

thestonefox: The problem with alphabetical order is it puts out the order of importance.

Unity
System
Malimbe
Zinnia
VRTK

Runtime/Haptics/HapticPulser.cs
thestonefox: what's the most efficient check here?

```
float clampedIntensity = Mathf.Clamp01(Intensity);
if (! Intensity.ApproxEquals(clampedIntensity)) {
  Intensity = clampedIntensity;
}
```

or just simply

```
if (Intensity < 0 || Intensity > 1)
{
    Intensity = Mathf.Clamp01(Intensity);
}
```

Runtime/Haptics/AudioClipHapticPulser.cs
bddckr: I was only suggesting that order inside the same root namespace, so it's still whatever order we want between those. We're not consistent so far anyway with this - there's plenty of using blocks not adhering to it which may be a result of me using auto-formatting as well as us doing refactoring and whole repo splits but then not going through all of these.

To me using statements are to make the compiler happy, I never look at them myself so I prefer automatic management of them.

Runtime/Haptics/HapticPulser.cs
bddckr: Let's use the former (without the space after the negation 😉) - the difference is negligible and that first solution reads easier as it says "this is what it should be, is it already? OK set it".

thestonefox: The former always performs a clamp operation even when it's not required. to me the latter is easier to understand.

if the number is less than 0 or greater than 1 then clamp between 0 and 1

you also don't have to do the clamp every time.

Runtime/Haptics/AudioClipHapticPulser.cs
thestonefox: i'd say we dont bother updating stuff like namespaces in these PRs (as the other tracking PR which is huge will have loads of these cases). I suggest we do a code format clean up at the end

Runtime/Haptics/HapticPulser.cs
bddckr: The most efficient solution is to clamp yourself, as you're already doing half of the job the Clamp call does by doing the if check here. But that's way too long.

Go with whatever you feel like.

PR #320, opened 2019-03-09
refactor(Rule): move rule observable lists into rule namespace

thestonefox:
The ObservableLists that are specific to the Rule namespace have
now been moved inside a Rule/Collection directory and the
Zinnia.Rule.Collection namespace for better housekeeping.


thestonefox:
Wait for https://github.com/ExtendRealityLtd/Zinnia.Unity/pull/317 to be merged first in case of conflicts (easier to fix here)

PR #325, opened 2019-03-11
feat(Event): update fields to properties and lists to observable

thestonefox:

The BehaviourEnabledObserver has also been re-written to utilise a
Coroutine instead of using InvokeRepeating.

The ObservableList has also had a new ReadOnlyCollection added that
allows for receiving the List collection at any time before Start
is called, however this list is not suitable for using with the
subscribed events or creating any reactive process based on the
List contents.

### DIFF
--- a/Runtime/Data/Collection/ObservableList.cs
+++ b/Runtime/Data/Collection/ObservableList.cs
@@ -55,11 +55,13 @@
         public int CurrentIndex { get; set; }
 
         /// <summary>
-        /// The collection to observe changes of.
+        /// The elements to observe changes of, accessible from components that *are* keeping in sync with the state of the collection by subscribing to the list mutation events. Alternatively use <see cref="NonSubscribableElements"/> instead.
         /// </summary>
-        public IReadOnlyList<TElement> ReadOnlyElements => wasStartCalled
-            ? (IReadOnlyList<TElement>)Elements
-            : Array.Empty<TElement>();
+        public IReadOnlyList<TElement> SubscribableElements => wasStartCalled ? (IReadOnlyList<TElement>)Elements : Array.Empty<TElement>();
+        /// <summary>
+        /// The elements to observe changes of, accessible from components that are *not* interested in keeping in sync with the state of the collection. Alternatively use <see cref="SubscribableElements"/> instead.
+        /// </summary>
+        public IReadOnlyList<TElement> NonSubscribableElements => Elements;
 
         /// <summary>
         /// The collection to observe changes of.
@@ -121,7 +123,7 @@
         [RequiresBehaviourState]
         public virtual void AddAt(TElement element, int index)
         {
-            if (ReadOnlyElements.Count == 0)
+            if (SubscribableElements.Count == 0)
             {
                 Add(element);
                 return;
@@ -153,7 +155,7 @@
         [RequiresBehaviourState]
         public virtual void SetAt(TElement element, int index)
         {
-            if (ReadOnlyElements.Count == 0)
+            if (SubscribableElements.Count == 0)
             {
                 return;
             }
@@ -214,13 +216,13 @@
         [RequiresBehaviourState]
         public virtual void RemoveAt(int index)
         {
-            if (ReadOnlyElements.Count == 0)
+            if (SubscribableElements.Count == 0)
             {
                 return;
             }
 
             index = Elements.GetWrappedAndClampedIndex(index);
-            TElement removedElement = ReadOnlyElements[index];
+            TElement removedElement = SubscribableElements[index];
             Elements.RemoveAt(index);
             EmitRemoveEvents(removedElement);
         }
@@ -241,7 +243,7 @@
         [RequiresBehaviourState]
         public virtual void Clear(bool removeFromFront)
         {
-            if (ReadOnlyElements.Count == 0)
+            if (SubscribableElements.Count == 0)
             {
                 return;
             }
@@ -251,7 +253,7 @@
                 Elements.Reverse();
             }
 
-            foreach (TElement element in ReadOnlyElements)
+            foreach (TElement element in SubscribableElements)
             {
                 ElementRemoved?.Invoke(element);
             }
@@ -264,14 +266,14 @@
         {
             wasStartCalled = true;
 
-            if (ReadOnlyElements == null)
+            if (SubscribableElements == null)
             {
                 return;
             }
 
-            for (int index = 0; index < ReadOnlyElements.Count; index++)
+            for (int index = 0; index < SubscribableElements.Count; index++)
             {
-                TElement element = ReadOnlyElements[index];
+                TElement element = SubscribableElements[index];
                 if (EqualityComparer<TElement>.Default.Equals(element, default))
                 {
                     continue;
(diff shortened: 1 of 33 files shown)


Runtime/Data/Collection/ObservableList.cs
```diff
@@ -55,11 +55,13 @@ public abstract class ObservableList<TElement, TEvent> : MonoBehaviour where TEv
         public int CurrentIndex { get; set; }
 
         /// <summary>
-        /// The collection to observe changes of.
+        /// The collection to observe changes of which can also be used for subscribing to the list mutation events but can only be accessed after <see cref="Start"/> has executed.
```
bddckr:
```suggestion
        /// The elements to observe changes of, accessible from components that *are* keeping in sync with the state of the collection by subscribing to the list mutation events. Alternatively use <see cref="NonSubscribableElements"/> instead.
```

Runtime/Data/Collection/ObservableList.cs
```diff
@@ -55,11 +55,13 @@ public abstract class ObservableList<TElement, TEvent> : MonoBehaviour where TEv
...
+        /// </summary>
+        public IReadOnlyList<TElement> SubscribableElements => wasStartCalled ? (IReadOnlyList<TElement>)Elements : Array.Empty<TElement>();
+        /// <summary>
+        /// The collection to observe changes of which can not be used for subscribing to the list mutation events but can be accessed at any point during the execution order.
```
bddckr:
```suggestion
        /// The elements to observe changes of, accessible from components that are *not* interested in keeping in sync with the state of the collection. Alternatively use <see cref="SubscribableElements"/> instead.
```

Runtime/Event/BehaviourEnabledObserver.cs
```diff
@@ -1,53 +1,136 @@
...
+    using Zinnia.Data.Collection;
 
     /// <summary>
     /// Emits an event once a list of <see cref="Behaviour"/>s all are <see cref="Behaviour.isActiveAndEnabled"/>.
     /// </summary>
     public class BehaviourEnabledObserver : MonoBehaviour
     {
         /// <summary>
-        /// The minimum time allowed to be passed to <see cref="MonoBehaviour.InvokeRepeating"/>.
+        /// The time between each <see cref="Behaviour.enabled"/> check before checking again.
```
bddckr:
```suggestion
        /// The time between each <see cref="Behaviour.enabled"/> check.
```

Runtime/Event/BehaviourEnabledObserver.cs
```diff
@@ -1,53 +1,136 @@
...
         public UnityEvent ActiveAndEnabled = new UnityEvent();
 
+        /// <summary>
+        /// A reference to the started routine.
+        /// </summary>
+        protected Coroutine behaviourCheckRoutine;
+        /// <summary>
+        /// Delays the <see cref="behaviourCheckRoutine"/> by <see cref="CheckDelay"/> seconds.
+        /// </summary>
+        protected WaitForSeconds checkDelayYield;
```
bddckr:
```suggestion
        protected WaitForSeconds checkDelayYieldInstruction;
```

Runtime/Event/BehaviourEnabledObserver.cs
```diff
@@ -1,53 +1,136 @@
...
+        /// <summary>
+        /// Cancels any running check of the <see cref="Behaviours"/> state.
+        /// </summary>
+        public virtual void EndCheck()
+        {
+            if (behaviourCheckRoutine != null)
+            {
+                StopCoroutine(behaviourCheckRoutine);
+            }
+            behaviourCheckRoutine = null;
```
bddckr:
```c#
            if (behaviourCheckRoutine == null)
            {
                return;
            }

            StopCoroutine(behaviourCheckRoutine);
            behaviourCheckRoutine = null;
```

Runtime/Event/BehaviourEnabledObserver.cs
```diff
@@ -1,53 +1,136 @@
...
+        {
+            float normalizedTime = 0;
+            checkDelayYield = new WaitForSeconds(CheckDelay);
+            while (normalizedTime <= 1f)
+            {
+                if (AreBehavioursEnabled())
+                {
+                    break;
+                }
+                normalizedTime += Time.deltaTime / MaximumRunTime;
```
bddckr: I don't understand this... Shouldn't the check just be like `float maximumTime = Time.time;` then `while (Time.time < maximumTime)`?

Runtime/Event/BehaviourEnabledObserver.cs
```diff
@@ -1,53 +1,136 @@
...
         /// <summary>
-        /// Checks whether all <see cref="behaviours"/> are <see cref="Behaviour.isActiveAndEnabled"/> and emits <see cref="ActiveAndEnabled"/> if they are.
+        /// Checks whether all <see cref="Behaviours"/> are <see cref="Behaviour.isActiveAndEnabled"/> and emits <see cref="ActiveAndEnabled"/> if they are.
         /// </summary>
-        protected virtual void Check()
+        /// <returns>Whether all <see cref="Behaviours"/> are active and enabled.</returns>
+        protected virtual bool AreBehavioursEnabled()
         {
-            if (!behaviours.TrueForAll(behaviour => behaviour.isActiveAndEnabled))
+            if (Behaviours == null || Behaviours.NonSubscribableElements.Count == 0)
```
bddckr: `NonSubscribableElements` here is correct, but brings up a point: Shouldn't (almost) all the (renamed) usages of `SubscribableElements` be `NonSubscribableElements` instead? I think we should ensure the correct ones are used... Maybe we just create a new ticket for it and come back to it another time.

Runtime/Event/BehaviourEnabledObserver.cs
thestonefox: you mean like

```
float endTime = Time.time + MaximumRunTime;
while (Time.time < endTime)
```

Runtime/Event/BehaviourEnabledObserver.cs
thestonefox: Using the `NonSubscribableElements` would also allow the tests that were updated to use [UnityTest] be able to go back to just using [Test] in many cases.

However, this is probably a large enough change to make it future work and not just add it to this PR.

Runtime/Event/BehaviourEnabledObserver.cs
```diff
@@ -1,53 +1,151 @@
...
+        {
+            float remainingRunTime = timeUntilCheckIsCancelled - Time.time;
+            timeUntilCheckIsCancelled = MaximumRunTime - remainingRunTime;
+        }
+
+        /// <summary>
+        /// Called after <see cref="CheckDelay"/> has been changed.
+        /// </summary>
+        [CalledAfterChangeOf(nameof(CheckDelay))]
+        protected virtual void OnAfterCheckDelayChange()
```
bddckr: Move this up so it's before `OnAfterMaximumRunTimeChange` to match the order of the properties' definition.

Runtime/Event/BehaviourEnabledObserver.cs
bddckr: I created https://github.com/ExtendRealityLtd/Zinnia.Unity/issues/326 for this.

Runtime/Event/BehaviourEnabledObserver.cs
thestonefox: 👍

PR #328, opened 2019-03-12
Feat/data changes

thestonefox:


Runtime/Data/Collection/GameObjectRelationObservableList.cs
```diff
@@ -0,0 +1,44 @@
...
+    using Malimbe.MemberClearanceMethod;
+    using Malimbe.XmlDocumentationAttribute;
+
+    /// <summary>
+    /// Allows observing changes to a <see cref="List{T}"/> of <see cref="Relation"/>s.
+    /// </summary>
+    public class GameObjectRelationObservableList : DefaultObservableList<GameObjectRelationObservableList.Relation, GameObjectRelationObservableList.UnityEvent>
+    {
+        /// <summary>
+        /// A relationship between a key <see cref="Relation"/> and it's associated value <see cref="Relation"/>.
```
bddckr:
```suggestion
        /// A relationship between a key <see cref="Relation"/> and its associated value <see cref="Relation"/>.
```

Runtime/Data/Collection/GameObjectRelationObservableList.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        [Serializable]
+        public class Relation
+        {
+            /// <summary>
+            /// The <see cref="GameObject"/> to use as the key for the relation.
```
bddckr:
```suggestion
            /// The <see cref="GameObject"/> acting as the key.
```

Runtime/Data/Collection/GameObjectRelationObservableList.cs
```diff
@@ -0,0 +1,44 @@
...
+            /// </summary>
+            [Serialized, Cleared]
+            [field: DocumentedByXml]
+            public GameObject Key { get; set; }
+            /// <summary>
+            /// The <see cref="GameObject"/> to use as the value for the relation.
```
bddckr:
```suggestion
            /// The <see cref="GameObject"/> acting as the value.
```

Runtime/Data/Type/TransformData.cs
```diff
@@ -3,82 +3,77 @@
...
         /// The state of whether the <see cref="TransformData"/> is valid.
         /// </summary>
-        public bool Valid
-        {
-            get
-            {
-                return (transform != null);
-            }
-        }
+        public bool Valid => Transform != null;
```
bddckr:
```suggestion
        public bool IsValid => Transform != null;
```
... while we're here.

Runtime/Visual/CameraColorOverlay.cs
```diff
@@ -234,7 +234,7 @@ protected virtual void CancelBlinkRoutine()
         /// <param name="cam">The <see cref="Camera"/> to apply onto.</param>
         protected virtual void PostRender(Camera cam)
         {
-            if (ValidCameras == null || !ValidCameras.cameras.Contains(cam))
+            if (CameraValidity == null || !CameraValidity.Accepts(cam))
```
bddckr:
```suggestion
            if (!CameraValidity.Accepts(cam))
```
`Accepts` is an extension method that already handles the rule being `null`.

While this change would make this rule usage consistent with all the other it does change the logic here, though:
With the suggested change a `null` rule would allow **any** camera, while currently we bail out treating it as no camera being valid instead.

What shall we do? 🤔

thestonefox: Whilst before it would never apply to all scene cameras if it was null, my gut is telling me that if you're using a rule and you don't apply a restriction then no restrictions should apply.

So I vote for removing the `== null` check

PR #329, opened 2019-03-12
feat(Cast): update fields to properties and lists to observable

thestonefox:

### DIFF
--- a/Runtime/Cast/FixedLineCast.cs
+++ b/Runtime/Cast/FixedLineCast.cs
@@ -1,10 +1,10 @@
 namespace Zinnia.Cast
 {
-    using Malimbe.PropertySerializationAttribute;
-    /*using Malimbe.PropertySetterMethod;*/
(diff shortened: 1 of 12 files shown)


Runtime/Cast/FixedLineCast.cs
```diff
@@ -24,41 +24,49 @@ public class FixedLineCast : StraightLineCast
...
+        /// <returns>The clamped value.</returns>
+        protected virtual float GetClampedCurrentLength()
         {
-            CurrentLength = Mathf.Clamp(CurrentLength, 0f, newValue);
+            return Mathf.Clamp(CurrentLength, 0f, MaximumLength);
         }
 
         /// <summary>
-        /// Handles changes to <see cref="CurrentLength"/>.
+        /// Called after <see cref="StraightLineCast.MaximumLength"/> has been changed.
```
bddckr: Replace the whole `summary` with `/// <inheritdoc />` instead.

PR #331, opened 2019-03-12
feat(Pointer): update fields to properties and lists to observable

thestonefox:

The ObjectPointer has also been split up so the Pointer Element
is now its own MonoBehaviour component to make it easier to
change pointer elements in the inspector.

### DIFF
--- a/Runtime/Pointer/ObjectPointer.cs
+++ b/Runtime/Pointer/ObjectPointer.cs
@@ -3,17 +3,20 @@
     using UnityEngine;
     using UnityEngine.Events;
     using System;
-    using Malimbe.BehaviourStateRequirementMethod;
+    using Malimbe.MemberClearanceMethod;
     using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
+    using Malimbe.BehaviourStateRequirementMethod;
     using Zinnia.Cast;
     using Zinnia.Data.Type;
     using Zinnia.Extension;
     using Zinnia.Visual;
+    using Zinnia.Process;
 
     /// <summary>
     /// Allows pointing at objects and notifies when a target is hit, continues to be hit or stops being hit by listening to a <see cref="PointsCast"/>.
     /// </summary>
-    public class ObjectPointer : MonoBehaviour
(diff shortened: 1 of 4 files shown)


Runtime/Pointer/ObjectPointer.cs
```diff
@@ -190,46 +152,26 @@ public enum Visibility
...
-        {
-            get;
-            protected set;
-        }
-
+        public bool IsActivated { get; protected set; }
+        /// <summary>
+        /// Whether the <see cref="ObjectPointer"/> is currently hovering over a target.
+        /// </summary>
+        /// <returns><see langword="true"/> if the <see cref="ObjectPointer"/> is currently hovering over a target, <see langword="false"/> otherwise.</returns>
```
bddckr:
```suggestion
```

Runtime/Pointer/ObjectPointer.cs
```diff
@@ -190,46 +152,26 @@ public enum Visibility
...
+        public bool IsHovering { get; protected set; }
         /// <summary>
         /// Reports hover duration of the <see cref="ObjectPointer"/> over the current target.
```
bddckr:
```suggestion
        /// How long the <see cref="ObjectPointer"/> is hovering over the current target.
```

Runtime/Pointer/PointerElement.cs
```diff
@@ -0,0 +1,50 @@
+﻿namespace Zinnia.Pointer
+{
+    using UnityEngine;
+    using Malimbe.MemberClearanceMethod;
+    using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
```
bddckr:
```suggestion
    using Malimbe.PropertySerializationAttribute;

```

Runtime/Pointer/ObjectPointer.cs
thestonefox: i've gone for `The duration that the <see cref="ObjectPointer"/> is hovering over the current target.`

bddckr: 👍

PR #338, opened 2019-03-15
feat(Attribute): add additional readonly options to Restricted

thestonefox:
The Restricted attribute now has a couple of additional options for
defining a read-only field. It now can be:

* ReadOnlyAlways: Always readonly
* ReadOnlyAtRunTime: Readonly when the app is playing
* ReadOnlyAtRunTimeAndEnabled: Readonly on playing and script enabled
* ReadOnlyAtRunTimeAndDisabled: Readonly on playing and script disabled

This offers a complete solution to times where readonly may be
required. Also, the `Muted` option has been moved to the top to make
things look neater.

### DIFF
--- a/Runtime/Data/Attribute/RestrictedAttribute.cs
+++ b/Runtime/Data/Attribute/RestrictedAttribute.cs
@@ -15,17 +15,25 @@
         public enum Restrictions
         {
             /// <summary>
+            /// The property is de-emphasized.
+            /// </summary>
+            Muted = 1 << 0,
+            /// <summary>
             /// The property is always read-only.
             /// </summary>
-            ReadOnlyAlways = 1 << 0,
+            ReadOnlyAlways = 1 << 1,
+            /// <summary>
+            /// The property is read-only when the application is playing.
+            /// </summary>
+            ReadOnlyAtRunTime = 1 << 2,
             /// <summary>
             /// The property is read-only when the application is playing and the component is enabled.
             /// </summary>
-            ReadOnlyWhenPlayingAndEnabled = 1 << 1,
+            ReadOnlyAtRunTimeAndEnabled = 1 << 4,
             /// <summary>
-            /// The property is de-emphasized.
+            /// The property is read-only when the application is playing and the component is disabled.
             /// </summary>
-            Muted = 1 << 2
+            ReadOnlyAtRunTimeAndDisabled = 1 << 8,
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -37,7 +45,7 @@ public enum Restrictions
         /// Draws the property in a specified restricted way.
         /// </summary>
         /// <param name="restrictions">The restriction options to apply.</param>
-        public RestrictedAttribute(Restrictions restrictions = Restrictions.ReadOnlyWhenPlayingAndEnabled | Restrictions.Muted)
+        public RestrictedAttribute(Restrictions restrictions = Restrictions.ReadOnlyAtRunTime | Restrictions.Muted)
```
bddckr: Shouldn't the default be `ReadOnlyAtRunTimeAndEnabled`?

Runtime/Data/Attribute/RestrictedAttribute.cs
```diff
@@ -14,18 +14,26 @@ public class RestrictedAttribute : PropertyAttribute
...
+            Muted = 1 << 0,
             /// <summary>
             /// The property is always read-only.
             /// </summary>
-            ReadOnlyAlways = 1 << 0,
+            ReadOnlyAlways = 1 << 1,
+            /// <summary>
+            /// The property is read-only when the application is playing.
+            /// </summary>
+            ReadOnlyAtRunTime = 1 << 2,
```
bddckr:
```suggestion
            ReadOnlyAtRuntime = 1 << 2,
```
https://en.wikipedia.org/wiki/Run_time

Not sure honestly, I just always read `runtime` and `design time`... (Also change the others if we decide to change this.)

thestonefox: Run Time is short for Running Time which is two separate words. Runtime to me is more of a noun than a verb (state)

Run time is the state in when the program is running

Runtime is the thing that is doing the running (e.g. Java Runtime)

if that makes sense?

bddckr: It does. 👍

Runtime/Data/Attribute/RestrictedAttribute.cs
thestonefox: Well that may not be the default required state. The default for our obs lists for example would be only readonly when it's running and disabled (because you can't make changes when disabled)

but then there are times where you may only want to make changes when enabled.

So in the end i thought it's probably best to just default handle both being considered the default, besides the existing use case of [Restricted] which is [InternalSetting] is for our facade settings, which shouldn't be changed at all at runtime. So i settled on ReadOnlyAtRunTime being the default (as it's going to be used pretty much everywhere in VRTK)

bddckr: 👍

PR #343, opened 2019-03-17
Feat/process improvements - resolves #307 - resolves #308

### LINKED ISSUE #307: MomentProcess.utilization should be framerate independent by using time instead of frame count

### LINKED ISSUE #308: MomentProcess.utilization start should be randomized between instances of processes
Especially for processes set up at design time they will all use the same counter start once playing the scene.

Instead of 100 processes running each frame we now have 100 processes running on every `n`th frame if they use the same utilization. This spike in CPU time should be prevented by randomizing the start of each process, so we get some distribution of the running processes.

For really advanced fine-tuning the start property should perhaps made available to users and their editor scripts, as one can imagine a script finding "the perfect" distribution for all objects known at design time. The random distribution may not do it for enough cases.

bddckr:


bddckr:
Re-requested a review as I added two more commits: One adds `ProcessNow`, the other adds (some) tests.

PR #346, opened 2019-03-17
chore(Association): explain why the test results in a warning log

bddckr:
One of the tests for `GameObjectsAssociationActivator` tests whether
an expected warning log message is actually logged. As this warning
log can be seen as a (partial) failure in the test by users running
the tests, the test has been adjusted to log additional information
before and after the expected warning log. These messages explain
that the warning is indeed expected and intended.

### DIFF
--- a/Tests/Editor/Association/GameObjectsAssociationActivatorTest.cs
+++ b/Tests/Editor/Association/GameObjectsAssociationActivatorTest.cs
@@ -96,8 +96,10 @@ namespace Test.Zinnia.Association
 
             subject.Associations.Add(associationMock);
 
+            Debug.Log("This test is expecting a warning to be logged next.");
             LogAssert.Expect(LogType.Warning, new Regex("multiple association"));
             subject.ManualAwake();
+            Debug.Log("Warning log recognized, the test is successful.");
         }
 
         [Test]


Tests/Editor/Association/GameObjectsAssociationActivatorTest.cs
```diff
@@ -96,8 +96,10 @@ public void AwakeLogsWarningForMultipleActiveGameObjects()
 
             subject.Associations.Add(associationMock);
 
+            Debug.Log("This test is expecting a warning to be logged next.");
```
thestonefox: What about instead of logging messages around the expect, what about using this to clear the log after the test has run?

https://docs.unity3d.com/ScriptReference/Debug.ClearDeveloperConsole.html

bddckr: I looked into this and I don't like it since it's an all-or-nothin approach. We want to hide that specific warning, not any others already in the console, nor any other logged during the run of the tests.

I tried setting our own log handler but it looks like there's no clean way to basically just filter what gets forwarded to Unity without introducing quite a bit of code just for that purpose.

thestonefox: This may also be helpful
https://answers.unity.com/questions/707636/clear-console-window.html

thestonefox: I'm fine with just logging this message then if that's the best of a bad situation.

PR #347, opened 2019-03-17
feat(CameraRig): add observable list for linked alias association

thestonefox:
The LinkedAliasAssociationCollection now has an ObservableList type for
holding a collection of linked alias associations.


thestonefox:
VRTK uses a LinkedAliasAssociationCollection in the TrackedAlias prefab in a List collection with manual ways of refreshing state if the list is updated. Better to have a way of using an Observable List.

PR #353, opened 2019-03-21
fix(Collision): provide access modifier on internal lists

thestonefox:
The CollisionIgnorer didn't specify the access modifier for the two
Collider lists being used internally which would mean they were set
to the default of `private` which prevents using them in any classes
that extend it.

They have now been set to `protected` so extending the class and using
these fields is possible.


thestonefox:
They're not public, they're protected. We still use protected fields elsewhere.

PR #355, opened 2019-03-21
feat(Collection): provide default value for Clear parameter

thestonefox:
The ObservableList Clear method now has a default value for the Clear
parameter set to false to just reduce the amount of times Clear(false)
is required, when simply Clear() can be used, which is the standard
operation.

### DIFF
--- a/Runtime/Data/Collection/ObservableList.cs
+++ b/Runtime/Data/Collection/ObservableList.cs
@@ -384,16 +384,16 @@
         /// <summary>
         /// Removes all elements from the collection.
         /// </summary>
-        /// <param name="removeFromFront">Whether to start the removal from the start or the end of the collection.</param>
+        /// <param name="removeFromEndToStart">Whether to reverse the collection when clearing so the removal goes from end to start.</param>
         [RequiresBehaviourState]
-        public virtual void Clear(bool removeFromFront)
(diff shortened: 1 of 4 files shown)


thestonefox:
I still think the Clear(var = false) is a good call so I deleted the other commit and just left that one in

Runtime/Data/Collection/ObservableList.cs
```diff
@@ -386,7 +386,7 @@ public virtual void RemoveAtCurrentIndex()
         /// </summary>
         /// <param name="removeFromFront">Whether to start the removal from the start or the end of the collection.</param>
         [RequiresBehaviourState]
-        public virtual void Clear(bool removeFromFront)
+        public virtual void Clear(bool removeFromFront = false)
```
bddckr: Shouldn't it be `true` instead? That saves performance and is in line with what `List<T>.Clear` does.

thestonefox: Yeah, but then on that I would say the variable should be renamed so the default is false and the internal check is positive

so like `bool reverseCollection = false` and then

```
if(reverseCollection) {
  reverse;
}
```

the param text could read

`Whether to reverse the collection when clearing so the removal goes from end to start.`

PR #363, opened 2019-03-26
Fix/pointers

thestonefox:


Runtime/Visual/PointsRenderer.cs
```diff
@@ -100,10 +100,11 @@ public virtual void RenderData(PointsData data)
 
             UpdateElement(data.Points, 0, false, Start);
             UpdateElement(data.Points, data.Points.Count - 1, false, End);
+            UpdateElement(data.Points, 0, true, RepeatedSegment);
 
-            for (int index = 0; index < segmentClones.Count; index++)
+            for (int index = 1; index <= segmentClones.Count; index++)
```
bddckr: All changes to this `for` loop can be reverted as the logic is the same (because the `index` is used in the loop body as `index - 1`.

thestonefox: If this index starts at 0 then it will do `index -1` and fail because 0 - 1 is not a valid index

unless i'm missing the point here?

bddckr: https://dotnetfiddle.net/SEamwq

bddckr: https://dotnetfiddle.net/EMJLk9

bddckr: I'm dumb - resolved.

PR #370, opened 2019-03-27
fix(Utility): error log when drawing field for component picker

bddckr:
The component picker draws an object field and then has to check how
that field was interacted with. Because the only public API offered
by Unity always creates an instance of the internal `ObjectSelector`
window that window could linger around in the editor, which resulted
in an error being logged whenever the game view was maximized. The
fix is to work around the internal logic of this API and instead use
intrinsic knowledge about how the internals works to figure out
whether a drop of a dragged object happened, or whether the picker
window was closed.


thestonefox:
Dragging and dropping no longer works

PR #371, opened 2019-03-29
Perf/more improvements

bddckr:


Runtime/Tracking/Modification/TransformPropertyApplier.cs
```diff
@@ -298,6 +297,13 @@ protected virtual Vector3 CalculatePositionWithOffset(Vector3 sourcePosition, Ve
         /// <param name="currentPosition">The current position to apply to the target.</param>
         protected virtual void ProcessTransform(TransformData source, TransformData target, Vector3 currentScale, Quaternion currentRotation, Vector3 currentPosition)
         {
+            if (ArePropertiesEqual(target.Position, currentPosition, target.Rotation, currentRotation, target.Scale, currentScale))
```
thestonefox: This actually works because the order is correct but it's technically inverted.

The `ArePropertiesEqual` method goes start/destination

`current` is the start and `target` is the destination

so this should read:

`ArePropertiesEqual(currentPosition, target.Position, currentRotation, target.Rotation, currentScale, target.Scale)`

Runtime/Tracking/Modification/TransformPropertyApplier.cs
```diff
@@ -152,7 +152,11 @@ public class UnityEvent : UnityEvent<EventData>
         /// <param name="source">The data to build the new source from.</param>
         public virtual void SetSource(GameObject source)
         {
-            Source = source != null ? new TransformData(source) : null;
+            sourceTransformData.Transform = source != null ? source.transform : null;
```
thestonefox: I'm not sure if this is the same logic.

`Source` is an injected object that then builds a representation of overrides within sourceTransformData (i think)

This is actually overriding the injected Source object with the internal representation of that object.

But then I'm struggling to see where `sourceTransformData` is even used. I'll need to investigate more

PR #375, opened 2019-04-05
Feat/rotation offset

thestonefox:


Runtime/Data/Operation/TransformEulerRotationMutator.cs
```diff
@@ -22,25 +34,87 @@ protected override float GetLocalAxisValue(int axis)
...
+        /// Returns the <see cref="Origin"/> position if a <see cref="Origin"/> is defined.
+        /// </summary>
+        /// <returns>The origin position.</returns>
+        protected virtual Vector3 GetOriginPosition()
+        {
+            return Origin != null ? Origin.transform.position : Vector3.zero;
+        }
+
+        /// <summary>
+        /// Applies the position of the <see cref="Origin"/> to the <see cref="Target"/> to ensure it rotates around the set origin.
```
bddckr:
```suggestion
        /// Applies the position of the <see cref="Origin"/> to the <see cref="TransformPropertyMutator.Target"/> to ensure it rotates around the set origin.
```

Runtime/Data/Operation/TransformPositionMutator.cs
```diff
@@ -2,20 +2,20 @@
 {
     using UnityEngine;
     using Malimbe.MemberClearanceMethod;
-    using Malimbe.PropertySerializationAttribute;
     using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
```
bddckr: Why was this changed? Now it's no longer in alphabetical order.

Runtime/Data/Type/Transformation/CollectionAggregator.cs
```diff
@@ -23,6 +23,23 @@ public abstract class CollectionAggregator<TInput, TOutput, TEvent, TCollection,
         [field: DocumentedByXml]
         public TCollection Collection { get; set; }
 
+        /// <summary>
+        /// Processes the current collection and sums the collection total.
```
bddckr:
```suggestion
        /// Processes the current collection and aggregates the collection total.
```

Runtime/Data/Type/Transformation/CollectionAggregator.cs
```diff
@@ -23,6 +23,23 @@ public abstract class CollectionAggregator<TInput, TOutput, TEvent, TCollection,
...
+        /// </summary>
+        /// <returns>The summed value of all values in the collection.</returns>
```
bddckr:
```suggestion
        /// <returns>The aggregated value of all values in the collection.</returns>
```

Runtime/Data/Type/Transformation/CollectionAggregator.cs
```diff
@@ -23,6 +23,23 @@ public abstract class CollectionAggregator<TInput, TOutput, TEvent, TCollection,
...
+        public virtual TOutput Transform()
+        {
+            return Transform(Collection.NonSubscribableElements[Collection.CurrentIndex]);
+        }
+
+        /// <summary>
+        /// Processes the current collection and sums the collection total.
```
bddckr:
```suggestion
        /// Processes the current collection and aggregates the collection total.
```

Runtime/Data/Type/Transformation/CollectionAggregator.cs
```diff
@@ -23,6 +23,23 @@ public abstract class CollectionAggregator<TInput, TOutput, TEvent, TCollection,
         [field: DocumentedByXml]
         public TCollection Collection { get; set; }
 
+        /// <summary>
+        /// Processes the current collection and sums the collection total.
+        /// </summary>
+        /// <returns>The summed value of all values in the collection.</returns>
+        public virtual TOutput Transform()
+        {
+            return Transform(Collection.NonSubscribableElements[Collection.CurrentIndex]);
```
bddckr: Just at a single index? Either this is wrong or the XML docs for this method (and `DoTransform`) should explain that this only operates on the `collection's current index` or something.

thestonefox: This is an implementation detail, it isn't required public knowledge about how it's performing the aggregation on the internal collection.

And the way it does that is to simply call `Transform(ObjectAtCurrentIndex)` 

which internally goes to the `Transform(Object)` method, which sets the given `Object` to the index of the collection at the `CurrentIndex`

So doing:

```
CurrentIndex = 1;
Transform(Object);

>>> does

Collection[CurrentIndex] = Object
ProcessCollection()
```

is the same as doing

```
CurrentIndex = 1
Transform()

>>> does

Transform()

Transform(ObjectAtCurrentIndex);
Collection[CurrentIndex] = Object
ProcessCollection()
```

Runtime/Data/Operation/TransformPositionMutator.cs
thestonefox: To match the style in most of the other scripts where namespaces flow with length to look neat. This should be resolved when we do a codebase wide operation applying a single rule of how we want the namespaces to work.

But for now, more often than not the namespaces aren't sorted alphabetically.

PR #382, opened 2019-04-27
feat(Pointer): rehandle existing pointer data

thestonefox:
There are occasions where the existing pointer data needs reprocessing
but the component requiring the rehandling does not contain pointer
data.

One case is when the PointerElement visibility changes, it requires
the existing pointer data to be reprocessed again.

This method provides a quick and simple way to reprocess the existing
pointer data.

### DIFF
--- a/Runtime/Pointer/ObjectPointer.cs
+++ b/Runtime/Pointer/ObjectPointer.cs
@@ -297,6 +297,15 @@
         }
 
         /// <summary>
+        /// Handles the previous provided <see cref="PointsCast.EventData"/> once more.
+        /// </summary>
+        [RequiresBehaviourState]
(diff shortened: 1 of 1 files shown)


Runtime/Pointer/ObjectPointer.cs
```diff
@@ -296,6 +296,15 @@ public virtual void HandleData(PointsCast.EventData data)
             UpdateRenderData();
         }
 
+        /// <summary>
+        /// Rehandles the previous provided <see cref="PointsCast.EventData"/>.
```
bddckr:
```suggestion
        /// Handles the previous provided <see cref="PointsCast.EventData"/> once more.
```

PR #384, opened 2019-04-29
fix(Tracking): remove gravity from rigidbody tests

thestonefox:
Gravity was being applied to the rigidbodies in the tracking
modifier tests causing them to sometimes be falling at the point
of the initial position/rotation assertion, which should have been
zero.

Gravity is now turned off on the rigidbody, which will prevent it
from falling and therefore the tests will pass.


thestonefox:
I definitely 100% ran the tests and they definitely 100% all pass!!

image

PR #379, opened 2019-04-17
fix AnyLayerRule

WingGao:

PR #387, opened 2019-05-07
fix(Rule): ensure layer rule matches correctly against layer mask

thestonefox:
The AnyLayerRule was not correctly matching against the specified
LayerMask and the tests were also incorrect and therefore not
highlighting the issue.

In the tests where it would set up the LayerMask by doing
`subject.LayerMask = LayerMask.NameToLayer("UI")` this would not
actually set the LayerMask property to `UI` instead it would set
it to `Default` and `Ignore Raycast`. The correct way to set the
LayerMask property is to use `LayerMask.GetMask("UI")` which then
will select the correct property and highlight the issue in the
AnyLayerRule.

The AnyLayerRule is then fixed by checking the LayerMask property
is equal to the existing LayerMask with the target layer pushed
on to it.

This makes all of the tests pass and also works by testing with
scene objects.

### DIFF
--- a/Runtime/Rule/AnyLayerRule.cs
+++ b/Runtime/Rule/AnyLayerRule.cs
@@ -19,7 +19,7 @@
         /// <inheritdoc />
         protected override bool Accepts(GameObject targetGameObject)
         {
-            return (targetGameObject.layer & LayerMask.value) != 0;
+            return (LayerMask & (1 << targetGameObject.layer)) != 0;
         }
     }
 }
--- a/Tests/Editor/Rule/AnyLayerRuleTest.cs
+++ b/Tests/Editor/Rule/AnyLayerRuleTest.cs
@@ -15,10 +15,7 @@ namespace Test.Zinnia.Rule
         [SetUp]
         public void SetUp()
         {
-            containingObject = new GameObject
-            {
-                layer = LayerMask.NameToLayer("UI")
-            };
+            containingObject = new GameObject();
             container = new RuleContainer();
             subject = containingObject.AddComponent<AnyLayerRule>();
             container.Interface = subject;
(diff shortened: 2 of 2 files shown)


Runtime/Rule/AnyLayerRule.cs
```diff
@@ -19,7 +19,7 @@ public class AnyLayerRule : GameObjectRule
         /// <inheritdoc />
         protected override bool Accepts(GameObject targetGameObject)
         {
-            return (targetGameObject.layer & LayerMask.value) != 0;
+            return LayerMask == (LayerMask | (1 << targetGameObject.layer));
```
bddckr:
```suggestion
            return (LayerMask & (1 << targetGameObject.layer)) != 0;
```
This should also do it and makes more sense, at least to me 😄: "Does this mask contain any layer of this game object, turned into a mask (by `1 << layer`?" Feel free to ignore this suggestion.

Do we want to put this into an extension method for `LayerMask` to solve once and for all or do we think this will be the only usage going forward?

thestonefox: I tried that (as it was in the other PR with a fix) but it still didn't work in actual practice for me

thestonefox: Also, it's the only usage right now, we should probably move to an extension if it is used elsewhere down the line. No point in making an extension for a single use case.

bddckr: Merge ahead then.

thestonefox: I do wonder why the alternative didn't work though. I figured that other person raised a PR with that as the fix, but then someone else said it didn't work for them in slack and it didn't work for me either.

bddckr: Maybe you were just missing parentheses or something? `(A & B) != 0` is definitely checking whether A "includes" B...

Alternatively remember that `GetMask` returns the mask (which is `1 << layerIndex`), `NameToLayer` returns just the layer's index (not a mask). Mask to int seems to be done via this: https://docs.unity3d.com/ScriptReference/LayerMask-value.html

thestonefox: NameToLayer definitely didn't work in the tests. It was setting the LayerMask to `Default` + `Ignore RayCast` instead of `UI`

this is the code I literally copied and pasted:

https://github.com/ExtendRealityLtd/Zinnia.Unity/pull/379/commits/691eccf65ec56ffbfe5c8baa73228274aad52a4c

For the initial attempt at fixing it (when I was trying it in the scene) and it didn't seem to work with that code.

But I've just noticed that code isn't what you posted above:

Other PR -> `return (targetGameObject.layer & (1 << LayerMask.value)) != 0;`

Your Code -> `return (LayerMask & (1 << targetGameObject.layer)) != 0;`

Original Code -> `return (targetGameObject.layer & LayerMask.value) != 0;`

So maybe your suggestion will work. I'll try it tomorrow and report back.

bddckr: Yeah as far as I understand Unity's docs the `Other PR` basically is confusing mask vs. layer index (i.e. `.value`).

Mask == `1 << layerIndex`, with `someMask.value` being the index? Docs are sparse on that one.

thestonefox: Ok so your original suggestion works and passes the new tests too.

PR #390, opened 2019-05-09
feat(Collision): ability to ignore tracked collision states

thestonefox:
The CollisionTracker now has an enum flag to determine which collision
states to process. Any ignored collision state simply will return early
in the OnCollision/Trigger method related to the state.

This gives the option of forcing any CollisionTracker to never process
a state if it's never required because at the moment it will always
at least run through the process of looking for CollisionNotifiers.

### DIFF
--- a/Runtime/Tracking/Collision/CollisionNotifier.cs
+++ b/Runtime/Tracking/Collision/CollisionNotifier.cs
@@ -137,11 +137,37 @@
         }
 
         /// <summary>
+        /// The states of a tracked collision.
+        /// </summary>
+        [Flags]
+        public enum CollisionStates
+        {
+            /// <summary>
+            /// When a new collision occurs.
+            /// </summary>
+            Enter = 1 << 0,
+            /// <summary>
+            /// When an existing collision continues to exist.
+            /// </summary>
+            Stay = 1 << 1,
+            /// <summary>
+            /// When an existing collision ends.
+            /// </summary>
+            Exit = 1 << 2
+        }
+
+        /// <summary>
         /// The types of collisions that events will be emitted for.
         /// </summary>
         [Serialized]
         [field: DocumentedByXml, UnityFlags]
         public CollisionTypes EmittedTypes { get; set; } = (CollisionTypes)(-1);
+        /// <summary>
+        /// The <see cref="CollisionStates"/> to process.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml, UnityFlags]
+        public CollisionStates StatesToProcess { get; set; } = (CollisionStates)(-1);
         /// <summary>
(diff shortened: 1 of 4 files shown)


thestonefox:
This was the idea I was discussing in slack about ignoring certain collision states.

thestonefox:
Yeah I guess it can also make sense on the superclass too. My only concern with making `Stay` off by default is then having to update all the prefabs :D

Runtime/Tracking/Collision/CollisionNotifier.cs
```diff
@@ -136,13 +136,39 @@ public enum CollisionTypes
             Trigger = 1 << 1
         }
 
+        /// <summary>
+        /// The states of a tracked collision
```
```suggestion
        /// The states of a tracked collision.
```

Runtime/Tracking/Collision/CollisionNotifier.cs
```diff
@@ -136,13 +136,39 @@ public enum CollisionTypes
...
+        }
+
         /// <summary>
         /// The types of collisions that events will be emitted for.
         /// </summary>
         [Serialized]
         [field: DocumentedByXml, UnityFlags]
         public CollisionTypes EmittedTypes { get; set; } = (CollisionTypes)(-1);
         /// <summary>
+        /// The Collision States to process.
```
```suggestion
        /// The <see cref="CollisionStates"/> to process.
```

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -9,31 +9,61 @@ public class CollisionTracker : CollisionNotifier
     {
         protected virtual void OnCollisionEnter(Collision collision)
         {
+            if ((StatesToProcess & CollisionStates.Enter) == 0)
```
bddckr: Remove all of these in this class, since the superclass already does this in the called `On*` method. This reduces the code duplication.

thestonefox: The reasons these are here is because thes `On` methods are part of the unity call and will always happen so we're going to do the event set and call to the inherited method if we don't bomb out early.

bddckr: You just call `OnCollisionStarted` here. `OnCollisionStarted` already does this `if`-check. Therefore you don't need the checks in this subclass.

thestonefox: it also calls `eventData.Set(this, false, collision, collision.collider)`

so it will set the event data, which is an extra operation but also is technically also now put the object in the incorrect state because the event data shouldn't be set if the operation shouldn't occur

bddckr: Oh that's what you meant. Yeah especially troublesome as `eventData` is defined on the superclass.

PR #392, opened 2019-05-15
feat(Tracking): terminate surface locator on certain collisions - fixes ExtendRealityLtd/VRTK#2012

thestonefox:
A new `LocatorTermination` Rule has been added to the SurfaceLocator
that allows a given rule to determine if the locator hits a target
which matches the rule to simply terminate the locator operation and
no longer continue to search for a valid surface.

This is in addition to the `TargetValidity` Rule which will ignore
invalid targets but continue searching for the next valid target. The
`LocatorTermination` Rule will prevent this search from continuing.

### DIFF
--- a/Runtime/Tracking/SurfaceLocator.cs
+++ b/Runtime/Tracking/SurfaceLocator.cs
@@ -79,6 +79,12 @@
         [field: Header("Restriction Settings"), DocumentedByXml]
         public RuleContainer TargetValidity { get; set; }
         /// <summary>
+        /// An optional <see cref="RuleContainer"/> to determine if the search for a valid surface should be terminated if the current found target matches the rule.
+        /// </summary>
+        [Serialized, Cleared]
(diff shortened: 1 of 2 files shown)


Runtime/Tracking/SurfaceLocator.cs
```diff
@@ -79,6 +79,12 @@ public class UnityEvent : UnityEvent<SurfaceData> { }
         [field: Header("Restriction Settings"), DocumentedByXml]
         public RuleContainer TargetValidity { get; set; }
         /// <summary>
+        /// An optional <see cref="RuleContainer"/> to determine if the surface locator should be terminated if the current found target matches the rule.
```
bddckr:
```suggestion
        /// An optional <see cref="RuleContainer"/> to determine if the search for a valid surface should be terminated if the current found target matches the rule.
```

PR #393, opened 2019-05-16
fix(Collision): stop collisions when either side is disabled - fixes ExtendRealityLtd/VRTK#2005

thestonefox:
There is an issue in Unity where when two objects are colliding will
create a collision started event, but whilst the objects are
intersecting, if one of the objects is disabled then Unity does not
consider this to stop the collision between the two objects so no
OnCollisionExit or OnTriggerExit occurs. However, if the disabled
object is re-enabled and is still in the same intersecting position
then Unity will consider this a new Enter event.

This causes a problem for the CollisionTracker as it means the state
of the tracker can become out of sync. If a collision starts twice
but no stop was called between, then the tracked collision state is
not valid. This would then have knock on effects in usages such as
the ActiveCollisionsContainer where a started collision is treated
as an active collision, but then if the object is disabled then
no stopped collision occurs and therefore that would still be
considered as an active collision.

To fix this issue and work around the shortcomings of Unity, a new
component is added at runtime to the GameObject that contains the
collider that is colliding with the CollisionTracker. This new
component then monitors the disabled state of the GameObject and upon
that GameObject becoming disabled will stop the collision manually
on the linked CollisionTracker. It will also then remove any traces
of the magically created component.

The CollisionTracker now also tracks any colliders it may be
intersecting with and therefore if the CollisionTracker is disabled
then any collider it may be linked to also has the Collision Stopped
event emitted for to ensure both sides of the collision adhere to the
disabled state.

This new functionality is optional and can be turned off by the `Stop
Collisions On Disable` property, which can only be set at edit time to
prevent fundamental state getting out of sync if it was changed at
runtime.

The `AddComponent` and `Destroy` are used over a pooling method because
these only occur on the `Enter/Exit` events so are not firing all the
time and therefore performance hits should be reduced. But also,
pooling would create a large set of GameObjects for each
CollisionTracker and they would all have to independently manage these
pools themselves resulting in a potential large number of pooled
GameObjects being created on scene start, which would have a potential
much larger impact on initial performance.

The tests have been updated to showcase the issue of disabling a
GameObject on either side of the collision now can successfully result
in the Stopped event being emitted.

### DIFF
--- a/Runtime/Tracking/Collision/CollisionTracker.cs
+++ b/Runtime/Tracking/Collision/CollisionTracker.cs
@@ -1,12 +1,58 @@
 ﻿namespace Zinnia.Tracking.Collision
 {
     using UnityEngine;
+    using System.Linq;
+    using System.Collections.Generic;
+    using Malimbe.MemberClearanceMethod;
+    using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
+    using Zinnia.Data.Attribute;
 
     /// <summary>
     /// Tracks collisions on the <see cref="GameObject"/> this component is on.
     /// </summary>
     public class CollisionTracker : CollisionNotifier
     {
+        /// <summary>
+        /// Causes collisions to stop if the <see cref="GameObject"/> on either side of the collision is disabled.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml, Restricted(RestrictedAttribute.Restrictions.ReadOnlyAtRunTime)]
+        public bool StopCollisionsOnDisable { get; protected set; } = true;
+
+        /// <summary>
+        /// A collection of current existing collisions.
+        /// </summary>
+        protected List<Collider> trackedCollisions = new List<Collider>();
+
+        /// <summary>
+        /// Stops the collision between this <see cref="CollisionTracker"/> and the given <see cref="Collider"/>. If there is still a physical intersection, then the collision will start again in the next physics frame.
+        /// </summary>
+        /// <param name="collider">The collider to stop the collision with.</param>
+        public virtual void StopCollision(Collider collider)
+        {
+            if ((StatesToProcess & CollisionStates.Exit) == 0)
+            {
+                return;
+            }
+
+            RemoveDisabledObserver(collider);
+            OnCollisionStopped(eventData.Set(this, collider.isTrigger, null, collider));
+        }
+
+        protected virtual void OnDisable()
+        {
+            if (!StopCollisionsOnDisable)
+            {
+                return;
+            }
+
+            foreach (Collider collider in trackedCollisions.ToArray())
+            {
+                StopCollision(collider);
+            }
+        }
+
         protected virtual void OnCollisionEnter(Collision collision)
         {
             if ((StatesToProcess & CollisionStates.Enter) == 0)
@@ -14,6 +60,7 @@
                 return;
             }
 
+            AddDisabledObserver(collision.collider);
             OnCollisionStarted(eventData.Set(this, false, collision, collision.collider));
         }
 
@@ -34,6 +81,7 @@
                 return;
             }
 
+            RemoveDisabledObserver(collision.collider);
             OnCollisionStopped(eventData.Set(this, false, collision, collision.collider));
         }
 
@@ -44,6 +92,7 @@
                 return;
             }
 
+            AddDisabledObserver(collider);
             OnCollisionStarted(eventData.Set(this, true, null, collider));
         }
 
@@ -64,7 +113,87 @@
                 return;
             }
 
+            RemoveDisabledObserver(collider);
             OnCollisionStopped(eventData.Set(this, true, null, collider));
+        }
+
+        /// <summary>
+        /// Adds a <see cref="CollisionTrackerDisabledObserver"/> to the <see cref="GameObject"/> that contains the <see cref="Collider"/> causing the collision.
+        /// </summary>
+        /// <param name="target">The target to add the <see cref="CollisionTrackerDisabledObserver"/> component to.</param>
+        protected virtual void AddDisabledObserver(Collider target)
+        {
+            if (target == null || !StopCollisionsOnDisable)
+            {
+                return;
+            }
+
+            trackedCollisions.Add(target);
+            CollisionTrackerDisabledObserver observer = target.gameObject.AddComponent<CollisionTrackerDisabledObserver>();
+            observer.Source = this;
+            observer.Target = target;
+        }
+
+        /// <summary>
+        /// Removes the <see cref="CollisionTrackerDisabledObserver"/> from the <see cref="GameObject"/> that contains the <see cref="Collider"/> ceasing the collision.
+        /// </summary>
+        /// <param name="target">The target to remove the <see cref="CollisionTrackerDisabledObserver"/> component from.</param>
+        protected virtual void RemoveDisabledObserver(Collider target)
+        {
+            if (target == null || !StopCollisionsOnDisable)
+            {
+                return;
+            }
+
+            foreach (CollisionTrackerDisabledObserver observer in target.gameObject.GetComponents<CollisionTrackerDisabledObserver>())
+            {
+                if (observer.Source == this && observer.Target == target)
+                {
+                    observer.Destroy();
+                    trackedCollisions.Remove(target);
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Observes the disabled state of any <see cref="GameObject"/> that the <see cref="CollisionTracker"/> is currently colliding with.
+    /// </summary>
+    public class CollisionTrackerDisabledObserver : MonoBehaviour
+    {
+        /// <summary>
+        /// The <see cref="CollisionTracker"/> that is causing the collision.
+        /// </summary>
+        [Serialized, Cleared]
+        [field: DocumentedByXml]
+        public CollisionTracker Source { get; set; }
+        /// <summary>
+        /// The <see cref="Collider"/> that is being collided with.
+        /// </summary>
+        [Serialized, Cleared]
+        [field: DocumentedByXml]
+        public Collider Target { get; set; }
+
+        /// <summary>
+        /// Whether <see cref="this"/> is being destroyed.
+        /// </summary>
+        protected bool isDestroyed;
+
+        /// <summary>
+        /// Destroys <see cref="this"/> from the scene.
+        /// </summary>
+        public virtual void Destroy()
+        {
+            isDestroyed = true;
+            Destroy(this);
+        }
+
+        protected virtual void OnDisable()
+        {
+            if (!isDestroyed)
+            {
+                Source.StopCollision(Target);
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -1,19 +1,66 @@
...
+        [field: DocumentedByXml, Restricted(RestrictedAttribute.Restrictions.ReadOnlyAtRunTime)]
+        public bool StopCollisionsOnDisable { get; protected set; } = true;
+
+        /// <summary>
+        /// A collection of current existing collisions.
+        /// </summary>
+        protected List<Collider> trackedCollisions = new List<Collider>();
+
+        /// <summary>
+        /// Stops the collision between this <see cref="CollisionTracker"/> and the given <see cref="Collider"/>. If there is still a physical intersection, then the collision will start again next physics frame.
```
bddckr:
```suggestion
        /// Stops the collision between this <see cref="CollisionTracker"/> and the given <see cref="Collider"/>. If there is still a physical intersection, then the collision will start again in the next physics frame.
```

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -1,19 +1,66 @@
...
+    using Malimbe.PropertySerializationAttribute;
+    using Zinnia.Data.Attribute;
 
     /// <summary>
     /// Tracks collisions on the <see cref="GameObject"/> this component is on.
     /// </summary>
     public class CollisionTracker : CollisionNotifier
     {
+        /// <summary>
+        /// Causes collisions to stop if the <see cref="GameObject"/> either side of the collision is disabled.
```
bddckr:
```suggestion
        /// Causes collisions to stop if the <see cref="GameObject"/> on either side of the collision is disabled.
```

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -1,19 +1,66 @@
...
+        }
+
         protected virtual void OnCollisionEnter(Collision collision)
         {
             if ((StatesToProcess & CollisionStates.Enter) == 0)
             {
                 return;
             }
 
+            AddTrackedNode(collision.collider);
```
bddckr: Move these calls (`AddTrackedNode` and `RemoveTrackedNode`) into the `On*` (e.g. this one's `OnCollisionStarted`) methods we call here. I.e. override the methods in this subclass.

Since the `On*` methods are our abstraction it feels like it makes more sense to do this additional logic in them rather than here, just for the sake of ensuring it's done when one would expect it to. This also ensure subclasses don't break that functionality when calling the `On*` methods themselves.

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -64,7 +113,87 @@ protected virtual void OnTriggerExit(Collider collider)
...
 
+            RemoveTrackedNode(collider);
             OnCollisionStopped(eventData.Set(this, true, null, collider));
         }
+
+        /// <summary>
+        /// Adds a <see cref="CollisionTrackerDisabledObserver"/> to the <see cref="GameObject"/> that contains the <see cref="Collider"/> causing the collision.
+        /// </summary>
+        /// <param name="target">The target to add the <see cref="CollisionTrackerDisabledObserver"/> component to.</param>
+        protected virtual void AddTrackedNode(Collider target)
```
bddckr:
```suggestion
        protected virtual void AddDisabledObserver(Collider target)
```

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -64,7 +113,87 @@ protected virtual void OnTriggerExit(Collider collider)
...
+        {
+            if (target == null || !StopCollisionsOnDisable)
+            {
+                return;
+            }
+
+            trackedCollisions.Add(target);
+            CollisionTrackerDisabledObserver node = target.gameObject.AddComponent<CollisionTrackerDisabledObserver>();
```
bddckr:
```suggestion
            CollisionTrackerDisabledObserver observer = target.gameObject.AddComponent<CollisionTrackerDisabledObserver>();
```

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -64,7 +113,87 @@ protected virtual void OnTriggerExit(Collider collider)
...
+                    trackedCollisions.Remove(target);
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Observes the disabled state of any <see cref="GameObject"/> that the <see cref="CollisionTracker"/> is currently colliding with.
+    /// </summary>
+    public class CollisionTrackerDisabledObserver : MonoBehaviour
```
bddckr: Put in standalone file instead?

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -64,7 +113,87 @@ protected virtual void OnTriggerExit(Collider collider)
...
+            CollisionTrackerDisabledObserver node = target.gameObject.AddComponent<CollisionTrackerDisabledObserver>();
+            node.Source = this;
+            node.Target = target;
+        }
+
+        /// <summary>
+        /// Removes the <see cref="CollisionTrackerDisabledObserver"/> from the <see cref="GameObject"/> that contains the <see cref="Collider"/> ceasing the collision.
+        /// </summary>
+        /// <param name="target">The target to remove the <see cref="CollisionTrackerDisabledObserver"/> component from.</param>
+        protected virtual void RemoveTrackedNode(Collider target)
```
bddckr:
```suggestion
        protected virtual void RemoveDisabledObserver(Collider target)
```

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -64,7 +113,87 @@ protected virtual void OnTriggerExit(Collider collider)
...
+        {
+            if (target == null || !StopCollisionsOnDisable)
+            {
+                return;
+            }
+
+            foreach (CollisionTrackerDisabledObserver node in target.gameObject.GetComponents<CollisionTrackerDisabledObserver>().ToArray())
```
bddckr:
```suggestion
            foreach (CollisionTrackerDisabledObserver observer in target.gameObject.GetComponents<CollisionTrackerDisabledObserver>())
```
Not sure why that `.ToArray` was used, but it seems unnecessary.

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -64,7 +113,87 @@ protected virtual void OnTriggerExit(Collider collider)
...
+
+        /// <summary>
+        /// Whether <see cref="this"/> is being destroyed.
+        /// </summary>
+        protected bool isDestroyed;
+
+        /// <summary>
+        /// Destroys <see cref="this"/> from the scene.
+        /// </summary>
+        public virtual void Destroy()
```
bddckr: Shouldn't this instead be `OnDestroyed`? What happens if a user destroys this component, even if it's because of destruction of a parent game object? This logic should still run, shouldn't it?

I may be missing some details around how this magic method works, I know you brought it up on Slack...

Runtime/Tracking/Collision/CollisionTracker.cs
```diff
@@ -64,7 +113,87 @@ protected virtual void OnTriggerExit(Collider collider)
...
+        {
+            isDestroyed = true;
+            Destroy(this);
+        }
+
+        protected virtual void OnDisable()
+        {
+            if (!isDestroyed)
```
bddckr: Why is this needed to be checked?

Runtime/Tracking/Collision/CollisionTracker.cs
thestonefox: Well this fix is actually fixing the issue Unity has with `OnCollisionXXX` and `OnTriggerXXX` what if there was a different implementation that extended the `CollisionNotifier`? You may not even need this work around.

I kinda feel like the `CollisionNotifier` is super generic and the `CollisionTracker` is our implementation of how Unity tracks collisions (badly).

Which would be my case for keeping these only in here and not dirtying the `CollisionNotifier`..... thoughts?

Runtime/Tracking/Collision/CollisionTracker.cs
thestonefox: `ToArray` was just a copy/paste from another foreach. I've deleted it as it isn't needed.

Runtime/Tracking/Collision/CollisionTracker.cs
thestonefox: Well if it's in a standalone file it can be dragged into the inspector, which we don't really want. It's basically the "magic" helper for this CollisionTracker.

Runtime/Tracking/Collision/CollisionTracker.cs
thestonefox: This isn't the Unity magic method which is `OnDestroy`

This is a custom method that destroys the observer component.

It's so the `CollisionTracker` can remove the component when the collision ends.

The reason it has the `isDestroyed` flag is so I know the component has been removed manually and therefore I don't want the `Source.StopCollision()` method firing when the component is disabled.

Because destroying the object will call `OnDisable` and then you'd get two stop messages..

If you destroy the object that this observer component is on, then `Destroy` isn't called and doesn't set `isDestroyed` so you *do* get the `OnDisable` called and therefore it does correctly call the `StopCollision`

Runtime/Tracking/Collision/CollisionTracker.cs
thestonefox: See the explanation above

Runtime/Tracking/Collision/CollisionTracker.cs
bddckr: Oh that's true. Anyone else using the `On*` methods should have proper ways to work with them right away, and if they don't they most likely are using this subclass anyway, so it wouldn't matter how we do it. As long as it's happening starting from Unity's magic callback methods that is, which this solution does.

Closing as resolved :)

Runtime/Tracking/Collision/CollisionTracker.cs
bddckr: `[AddComponentMenu("")]` should hide it from there! I think we even use that elsewhere already :)

thestonefox: It hides it from the drop down but doesn't prevent you from dragging that component into the inspector window, which we still dont want.

Runtime/Tracking/Collision/CollisionTracker.cs
bddckr: I looked into this some more and there is indeed no way to say "OnDisable, but only when *not* destroyed". Too bad Unity always throws in some weird specific implementation detail.

Runtime/Tracking/Collision/CollisionTracker.cs
bddckr: 🤷‍♂ There's only so much I'd do to prevent obvious mistakes.

I just don't like relying on this Unity behavior as it seems undocumented, or at least not documented as a "feature" that is intended this way..

PR #395, opened 2019-05-18
fix(Collision): perform equality check on collider and forward source

thestonefox:
The CollisionNotifier.EventData was performing the equality check on
the ForwardSource of the collision and the containing Transform of
the collider being intersected. This actually can report invalid
equality data as two compound colliders being touched by the same
ForwardSource would report as being the same Collision Data which
they're actually not.

The fix is to just do the equality check directly on the Collider and
the ForwardSource.

The HashCode of the EventData should also be a combination of the
two equality checked objects to ensure its uniqueness.

### DIFF
--- a/Runtime/Tracking/Collision/CollisionNotifier.cs
+++ b/Runtime/Tracking/Collision/CollisionNotifier.cs
@@ -80,7 +80,7 @@
                     return true;
                 }
 
-                return Equals(ColliderData.GetContainingTransform(), other.ColliderData.GetContainingTransform()) && ((ForwardSource == null && other.ForwardSource == null) || Equals(ForwardSource, other.ForwardSource));
+                return Equals(ForwardSource, other.ForwardSource) && Equals(ColliderData, other.ColliderData);
             }
 
             /// <inheritdoc />
@@ -96,18 +96,13 @@
                     return true;
                 }
 
-                if (obj.GetType() != GetType())
-                {
-                    return false;
-                }
-
-                return Equals((EventData)obj);
+                return obj is EventData other && Equals(other);
             }
 
             /// <inheritdoc />
             public override int GetHashCode()
             {
-                return ColliderData.GetContainingTransform().GetHashCode();
+                return ((ForwardSource != null ? ForwardSource.GetHashCode() : 0) * 397) ^ (ColliderData != null ? ColliderData.GetHashCode() : 0);
             }
 
             public static bool operator ==(EventData left, EventData right) => Equals(left, right);


Runtime/Tracking/Collision/CollisionNotifier.cs
```diff
@@ -107,7 +107,7 @@ public override bool Equals(object obj)
             /// <inheritdoc />
             public override int GetHashCode()
             {
-                return ColliderData.GetContainingTransform().GetHashCode();
+                return ColliderData.GetHashCode() + (ForwardSource != null ? ForwardSource.GetHashCode() : 0);
```
```suggestion
                return ((ForwardSource != null ? ForwardSource.GetHashCode() : 0) * 397) ^ (ColliderData != null ? ColliderData.GetHashCode() : 0);
```
(I used ReSharper's auto-generation for this.)

Runtime/Tracking/Collision/CollisionNotifier.cs
```diff
@@ -80,7 +80,7 @@ public bool Equals(EventData other)
                     return true;
                 }
 
-                return Equals(ColliderData.GetContainingTransform(), other.ColliderData.GetContainingTransform()) && ((ForwardSource == null && other.ForwardSource == null) || Equals(ForwardSource, other.ForwardSource));
+                return Equals(ColliderData, other.ColliderData) && ((ForwardSource == null && other.ForwardSource == null) || Equals(ForwardSource, other.ForwardSource));
```
```suggestion
                return Equals(ForwardSource, other.ForwardSource) && Equals(ColliderData, other.ColliderData);
```
The `ReferenceEquals` above and the way `Equals` work already handle `null`-cases.

Runtime/Tracking/Collision/CollisionNotifier.cs
```diff
@@ -80,7 +80,7 @@ public bool Equals(EventData other)
                     return true;
                 }
 
-                return Equals(ColliderData.GetContainingTransform(), other.ColliderData.GetContainingTransform()) && ((ForwardSource == null && other.ForwardSource == null) || Equals(ForwardSource, other.ForwardSource));
+                return Equals(ColliderData, other.ColliderData) && ((ForwardSource == null && other.ForwardSource == null) || Equals(ForwardSource, other.ForwardSource));
             }
 
             /// <inheritdoc />
```
```suggestion
            /// <inheritdoc />
            public override bool Equals(object obj)
            {
                if (ReferenceEquals(null, obj))
                {
                    return false;
                }

                if (ReferenceEquals(this, obj))
                {
                    return true;
                }

                return obj is EventData other && Equals(other);
            }
```
Note the last line: We want to allow subclassing this and thus equality needs to be overridable properly, which this handles.

thestonefox: are you saying here replace
```
public override bool Equals(object obj)
{
    if (ReferenceEquals(null, obj))
    {
        return false;
    }

    if (ReferenceEquals(this, obj))
    {
        return true;
    }

    if (obj.GetType() != GetType())
    {
        return false;
    }

    return Equals((EventData)obj);
}
```

with

```
public override bool Equals(object obj)
{
  if (ReferenceEquals(null, obj))
  {
    return false;
  }
  if (ReferenceEquals(this, obj))
  {
    return true;
  }
  return obj is EventData other && Equals(other);
}
```

bddckr: Yes :) (Newlines or not to separate the `if`s I'm fine with 😄)

thestonefox: The newlines were removed due to some bad copy/pasting :)

PR #396, opened 2019-05-23
fix(Collision): prevent notifier collection enumeration multiple times

thestonefox:
The CollisionNotifier reuses a collection for the found collision
notifiers to remove any garabge, however, this caused an issue if the
OnCollisionXXX methods were called again by anything during the process
as any additional call would attempt to re-load in any found collision
notifiers thus mutating the collection and causing the initial
`foreach` to fail due to the collection being changed.

As there is no need to allow the reprocessing of the notifier
collection as all notifiers should get processed on the first pass,
then the fix is to simply exit the methods early if they are called
again from the within the `foreach` instruction.

The cached collection has now been split up into 3 distinct collections
for each of the collision types ecause there could be occasions where
different OnCollision methods execute other OnCollision methods
e.g. OnCollisionStarted could via an event emit an OnCollisionStopped
and this would then either cause a mutation and therefore a failure
or prevent the correct OnCollision method being called on the linked
notifiers.

### DIFF
--- a/Runtime/Tracking/Collision/CollisionNotifier.cs
+++ b/Runtime/Tracking/Collision/CollisionNotifier.cs
@@ -191,9 +191,30 @@
         /// </summary>
         protected readonly EventData eventData = new EventData();
         /// <summary>
-        /// A reused instance to use when looking up <see cref="CollisionNotifier"/> components.
+        /// A reused instance to use when looking up <see cref="CollisionNotifier"/> components on start.
         /// </summary>
-        protected readonly List<CollisionNotifier> collisionNotifiers = new List<CollisionNotifier>();
+        protected readonly List<CollisionNotifier> startCollisionNotifiers = new List<CollisionNotifier>();
+        /// <summary>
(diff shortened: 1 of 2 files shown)


Runtime/Tracking/Collision/CollisionNotifier.cs
```diff
@@ -191,9 +191,30 @@ public enum CollisionStates
...
+        protected readonly List<CollisionNotifier> changeCollisionNotifiers = new List<CollisionNotifier>();
+        /// <summary>
+        /// A reused instance to use when looking up <see cref="CollisionNotifier"/> components on stop.
+        /// </summary>
+        protected readonly List<CollisionNotifier> stopCollisionNotifiers = new List<CollisionNotifier>();
+
+        /// <summary>
+        /// Whether the <see cref="collisionNotifiers"/> collection is being processed on start.
+        /// </summary>
+        protected bool processingStartNotifierCollection;
```
```suggestion
        protected bool isProcessingStartNotifierCollection;
```
Same for the other two.

PR #397, opened 2019-06-04
Fix/interaction perf

thestonefox:


Runtime/Extension/ColliderExtensions.cs
```diff
@@ -19,8 +19,27 @@ public static Transform GetContainingTransform(this Collider collider)
...
+        /// Gets the parent <see cref="Rigidbody"/> for the given <see cref="Collider"/> even if the <see cref="GameObject"/> is disabled.
+        /// </summary>
+        /// <param name="collider">The <see cref="Collider"/> to check against.</param>
+        /// <returns>The parent <see cref="Rigidbody"/>.</returns>
+        public static Rigidbody GetAttachedRigidbody(this Collider collider)
+        {
+            Rigidbody attachedRigidbody = collider.attachedRigidbody;
+            if (!collider.gameObject.activeInHierarchy)
+            {
+                foreach (Rigidbody foundRigidbody in collider.GetComponentsInParent<Rigidbody>(true))
```
bddckr: This allocates a new list each time to return to us. Instead we should do what we do elsewhere: Create our own list and reuse it in the call to `GetComponentsInParent` (there's a `bool, List<T>` overload).

Runtime/Tracking/Collision/Active/ActiveCollisionsContainer.cs
```diff
@@ -196,15 +224,71 @@ protected virtual CollisionNotifier.EventData CloneEventData(CollisionNotifier.E
...
         {
-            Transform containingTransform = collisionData.ColliderData.GetContainingTransform();
             return containingTransform != null && CollisionValidity.Accepts(containingTransform.gameObject);
         }
 
+        /// <summary>
+        /// Adds the given data to the <see cref="Elements"/> collection if it is not already present.
+        /// </summary>
+        /// <param name="data">The data to add.</param>
+        protected virtual void AddUniqueElement(CollisionNotifier.EventData data)
```
bddckr: Perhaps this whole method should be removed and `Elements` should become `HashSet<CollisionNotifier.EventData>` instead? Better perf that way.

Runtime/Tracking/Collision/Active/ActiveCollisionsContainer.cs
```diff
@@ -196,15 +224,71 @@ protected virtual CollisionNotifier.EventData CloneEventData(CollisionNotifier.E
...
+            }
+        }
+
+        /// <summary>
+        /// Determines if the containing <see cref="Transform"/> is already being collided with by a different nested <see cref="Collider"/>.
+        /// </summary>
+        /// <param name="collisionData">The collision data containing the <see cref="Collider"/> to check.</param>
+        /// <param name="currentCollisionContainingTransform">The containing <see cref="Transform"/> for the current collider.</param>
+        /// <returns>Whether there are any existing collisions occuring with the containing <see cref="Transform"/>.</returns>
+        protected virtual bool AreExistingCollisions(CollisionNotifier.EventData collisionData, Transform currentCollisionContainingTransform)
```
bddckr: This method does way more than doing a simple lookup/check as the name implies. It feels worth to rename it to explicitly describe the state mutation it's doing and keep the return value documented via the XML docs. Suggestion: `AddCollisionData`

Runtime/Tracking/Collision/Active/ActiveCollisionsContainer.cs
thestonefox: The `Elements` are required to be ordered as they are emitted in the event for processing elsewhere.

Making it a HashSet would mean something then would need to turn it back into a List for sorting down the line.

bddckr: `SortedSet` then!

Runtime/Tracking/Collision/Active/ActiveCollisionsContainer.cs
thestonefox: It's not actually adding collision data persay.

It's checking if there are any existing collisions for the given ContainingTransfomr and if there are no collisions at all for the containing transform then it returns false.

If there are collisions for the containing transform that match directly for the current collision then it returns true.

The rest of it is just maintaining the `Elements` array so the latest collision for the containing transform is stored.

Perhaps instead the `AreExistingCollisions` can stay as is, but the mutation of `Elements` can be moved into its own method (e.g. UpdateContainerCollisions)

Then the call could be something like

```
if (AreExistingCollisions(collisionData, currentCollisionContainingTransform)) {
 UpdateContainerCollisions(params)
} 
else {
...exiting below code
}
```

Runtime/Tracking/Collision/Active/ActiveCollisionsContainer.cs
thestonefox: I'd say we defer that change for another time as the List is processed in multiple places already meaning this change would get much larger.

Plus right now the Contains check is only done on Start and Stop collision so it's not firing every frame, so the perf hit of the Contains check isn't massive right now.

bddckr: I don't think the public API would change much but 🤷‍♂ fine as is.

Runtime/Tracking/Collision/Active/ActiveCollisionsContainer.cs
bddckr: 👍

bddckr:
Btw I'm ignoring the fact that the Dictionary with List values creates garbage repeatedly. The only good way to solve that is by pooling, but I don't want to introduce that to the codebase just for this purpose if we can help it.

thestonefox:
The profiler isn't showing any garbage generated to me. I'm just getting 

image

bddckr:
You'll have to hit the scenario of the list creation a few times to see it. I.e. the containing transform needs to be different... It's definitely creating garbage when we create new lists on the fly 😄 It will be fine I'm sure, only happens on start/stop, not on `Stay`.

Runtime/Tracking/Collision/Active/ActiveCollisionsContainer.cs
thestonefox: Things like `AddRange` and accessing with `[index]` don't work with a `SortedSet` which are used elsewhere

Runtime/Extension/ColliderExtensions.cs
thestonefox: Is it acceptable here to create a `private static List<Rigidbody>` in the class definition?

Or is making this static going to cause issues as it will become global?

I guess the `GetComponentsInParent` call will always override the list contents each time it is populated though?

bddckr: `static` will be fine and reduce the memory need down to a **single** instance of a list to use. (This is where Unity's restriction to only ever work on the main thread in a single-threaded fashion for all of their callbacks and API allows us to do such things.)

Runtime/Tracking/Collision/Active/ActiveCollisionsContainer.cs
thestonefox: lol I can't actually do what I suggested because `true` is returned twice but different logic happens so separating it out isn't going to help

I need the logic to be:

* Is there a collision for the containing transform already? no then return false
* There is a containing transform collision, but is there a collision for the current collision data? yes, then return true
* If there is a containing transform collision but a new nested collision then update the `Elements` with this new collision and remove any old collisions, then return true

If i split out that 3rd point then the logic will all be wrong.

I could still move the Remove/Add logic to a sub method to make it easier to read, but that means then the `AreExistingCollisions` method is still internally mutating that `Elements` collection so we're back to "the name doesn't really say what's fully going on"

PR #399, opened 2019-06-11
feat(Type): add collapsible UnityEvent drawer

thestonefox:
A new custom PropertyDrawer that overrides the UnityEventDrawer has
been added to allow the ability to wrap a UnityEvent inspector control
in a foldout so the UnityEvent control can be collapsed.

Reflection is required to override the default UnityEventDrawer
inspector drawing ue to the way Unity creates the UnityEventDrawer
internally.

### DIFF
--- a/Editor/Data/Type/CollapsibleUnityEventDrawer.cs
+++ b/Editor/Data/Type/CollapsibleUnityEventDrawer.cs
@@ -1,0 +1,103 @@
+﻿namespace Zinnia.Data.Type
+{
+    using UnityEngine;
+    using UnityEditor;
+    using UnityEditorInternal;
+    using UnityEngine.Events;
+    using UnityEngine.EventSystems;
+    using System.Linq;
+    using System.Reflection;
+    using System.Collections;
+
+    [CustomPropertyDrawer(typeof(UnityEventBase), true)]
+    [CustomPropertyDrawer(typeof(UnityEvent), true)]
+    [CustomPropertyDrawer(typeof(UnityEvent<>), true)]
+    [CustomPropertyDrawer(typeof(UnityEvent<BaseEventData>), true)]
+    public class CollapsibleUnityEventDrawer : UnityEventDrawer
+    {
+        protected const float headerHeight = 20f;
+        protected const float padding = 6f;
+        protected const float heightOffset = 2f;
+        protected readonly GUIStyle headerBackground = new GUIStyle("RL Header");
+        protected bool wasBaseOnGuiCalled = false;
+
+        [InitializeOnLoadMethod]
+        public static void ReplaceDefaultDrawer()
+        {
+            System.Type utilityType = System.Type.GetType("UnityEditor.ScriptAttributeUtility, UnityEditor");
+            MethodInfo buildMethod = utilityType.GetMethod(
+                "BuildDrawerTypeForTypeDictionary",
+                BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static,
+                null,
+                System.Type.EmptyTypes,
+                null);
+            FieldInfo dictionaryField = utilityType.GetField(
+                "s_DrawerTypeForType",
+                BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static);
+            FieldInfo drawerField = utilityType
+                .GetNestedType("DrawerKeySet", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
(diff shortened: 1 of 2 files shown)


Editor/Data/Type/CollapsibleUnityEventDrawer.cs
```diff
@@ -0,0 +1,103 @@
+﻿namespace Zinnia.Data.Type
+{
+    using System.Collections;
```
bddckr: Don't we put Unity stuff first normally?

Editor/Data/Type/CollapsibleUnityEventDrawer.cs
```diff
@@ -0,0 +1,103 @@
...
+    using UnityEditorInternal;
+    using UnityEngine;
+    using UnityEngine.Events;
+    using UnityEngine.EventSystems;
+
+    [CustomPropertyDrawer(typeof(UnityEventBase), true)]
+    [CustomPropertyDrawer(typeof(UnityEvent), true)]
+    [CustomPropertyDrawer(typeof(UnityEvent<>), true)]
+    [CustomPropertyDrawer(typeof(UnityEvent<BaseEventData>), true)]
+    internal class CollapsibleUnityEventDrawer : UnityEventDrawer
```
bddckr:
```suggestion
    public class CollapsibleUnityEventDrawer : UnityEventDrawer
```

Editor/Data/Type/CollapsibleUnityEventDrawer.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        protected const float headerHeight = 20f;
+        protected const float padding = 6f;
+        protected const float heightOffset = 2f;
+        protected readonly GUIStyle headerBackground = "RL Header";
```
bddckr:
```suggestion
        protected const GUIStyle headerBackground = "RL Header";
```
If that doesn't work I'd suggest this instead:
```suggestion
        protected static readonly GUIStyle headerBackground = "RL Header";
```

Editor/Data/Type/CollapsibleUnityEventDrawer.cs
```diff
@@ -0,0 +1,103 @@
...
+            else
+            {
+                DrawHeader(position);
+            }
+        }
+
+        public override float GetPropertyHeight(SerializedProperty property, GUIContent label) =>
+            property.isExpanded
+                ? base.GetPropertyHeight(property, label)
+                : EditorGUIUtility.singleLineHeight + EditorGUIUtility.standardVerticalSpacing * 2f;
```
bddckr:
```suggestion
                : EditorGUIUtility.singleLineHeight + heightOffset * 2f;
```
At least I believe so 🤔

Editor/Data/Type/CollapsibleUnityEventDrawer.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (Event.current.type == EventType.Repaint)
+            {
+                headerBackground.Draw(position, false, false, false, false);
+            }
+
+            Rect headerRect = new Rect(position.x, position.y, position.width, headerHeight);
+            headerRect.xMin += padding;
+            headerRect.xMax -= padding;
+            headerRect.height -= heightOffset;
+            headerRect.y += 1f;
```
bddckr:
```suggestion
            headerRect.y += heightOffset / 2f;
```
Same here: I think that's the actual source of that value.

Editor/Data/Type/CollapsibleUnityEventDrawer.cs
thestonefox: This in fact should be 
`protected static readonly GUIStyle headerBackground = new GUIStyle("RL Header");`

I think if you don't creata a new GUIStyle then it just references the current session style in the Unity editor and if you change anything you end up changing the style editorwide.

thestonefox: actually making it static breaks things too.... shruggie emoji

Editor/Data/Type/CollapsibleUnityEventDrawer.cs
thestonefox: still renders ok

Editor/Data/Type/CollapsibleUnityEventDrawer.cs
bddckr: Leaving as is then haha 🤷‍♂

PR #404, opened 2019-06-16
Fix/obs list fixes

thestonefox:


Editor/Data/Type/DefaultObservableListDrawer.cs
```diff
@@ -45,7 +47,14 @@ public override void OnGUI(Rect position, SerializedProperty property, GUIConten
                 using (new EditorGUI.IndentLevelScope())
                 {
                     dynamic list = propertyObject;
-                    bool listIsEmpty = list.NonSubscribableElements.Count == 0;
+                    dynamic elements = list.NonSubscribableElements;
+                    bool listIsEmpty = elements.Count == 0;
```
bddckr:
```suggestion
                    bool isListEmpty = elements.Count == 0;
```
I overlooked this one in the previous PR.

Editor/Data/Type/DefaultObservableListDrawer.cs
```diff
@@ -70,6 +79,7 @@ public override void OnGUI(Rect position, SerializedProperty property, GUIConten
                                 using (EditorGUI.ChangeCheckScope check = new EditorGUI.ChangeCheckScope())
                                 {
                                     dynamic elementValue = EditorGUILayout.ObjectField(currentElement, elementType, true);
+                                    //dynamic elementValue = EditorGUILayout.PropertyField(currentElement);
```
bddckr: Delete this 😄

Editor/Data/Type/DefaultObservableListDrawer.cs
```diff
@@ -24,6 +25,7 @@ public override void OnGUI(Rect position, SerializedProperty property, GUIConten
         {
             label.tooltip = EditorHelper.GetTooltipAttribute(fieldInfo)?.tooltip ?? string.Empty;
             EditorGUILayout.PropertyField(property, label, true);
+            string fieldLabel = label.text;
```
bddckr: Can we delete this and just inline `label.text` in the single usage of this variable below?

thestonefox: for some reason when the `Foldout` runs, the scope of `label` changes to the label of the Foldout (which is `Elements`)

so unless I store it prior to that, I lose access to `label.text`

thestonefox: https://answers.unity.com/questions/1408104/propertydrawerongui-label-parameter-contents-chang.html something something heap allocation

PR #405, opened 2019-06-16
fix(Data): utilize SerializedObject data to draw list elements

thestonefox:
There is an issue with manually building the list out of the
instantiated list object data as it will never pass through Unity's
serializer and therefore any changes will not be serialized to disk
meaning the control doesn't actually persist changes when the
scene is saved.

This fix removes all of the custom code and simply gets a
SerializedObject of the referenced property and uses the Unity
built in PropertyField to render the field which will ensure the
property changes are serialized to disk.

This does have the drawback that the rendered property drawer does
not utilize the custom ObservableListEditor which handles manually
firing the Observable List events when changes occur. This means
any changes to the Elements list in a reference property (i.e. using
this custom property drawer) will not emit any of the change events
found on the Observable List component.

Whilst this isn't an ideal solution, it isn't a breaking issue as
the underlying Observable List component still will emit the events
when changed in the inspector so any such requirements would just
simply need the actual Observable List component to be worked on
at runtime in the editor and this Property Drawer is simply considered
a nice shortcut helper for setting up references quickly in the editor.

### DIFF
--- a/Editor/Data/Type/DefaultObservableListDrawer.cs
+++ b/Editor/Data/Type/DefaultObservableListDrawer.cs
@@ -2,139 +2,31 @@
 {
     using UnityEngine;
     using UnityEditor;
-    using System;
-    using Zinnia.Utility;
     using Zinnia.Data.Collection.List;
 
     [CustomPropertyDrawer(typeof(ObservableList), true)]
     public class DefaultObservableListDrawer : PropertyDrawer
     {
-        protected const string collectionLabel = "Elements";
-        protected const string collectionElementLabel = "Element {0}";
-        protected const string collectionEmptyLabel = "List is Empty";
-        protected const string invalidElementTypeMessage = "Element of Type `{0}` is not supported. Click the reference field `{1}` above and edit the `Elements` collection directly on the source component.";
-        protected const float footerSpacing = 2f;
-        protected const float objectPickerPadding = 4f;
-        protected const int indentResetLevel = -2;
-        protected readonly GUIStyle buttonStyle = new GUIStyle("RL Footer");
-        protected readonly GUIContent objectPickerIcon = EditorGUIUtility.TrIconContent("In ObjectField");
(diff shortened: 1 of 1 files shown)


Editor/Data/Type/DefaultObservableListDrawer.cs
```diff
@@ -2,139 +2,30 @@
...
+        protected readonly GUIContent propertyReferenceLabel = new GUIContent("Elements");
         public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
         {
-            label.tooltip = EditorHelper.GetTooltipAttribute(fieldInfo)?.tooltip ?? string.Empty;
             EditorGUILayout.PropertyField(property, label, true);
-            string fieldLabel = label.text;
+
             object propertyObject = fieldInfo.GetValue(property.serializedObject.targetObject);
-            if (propertyObject == null)
+            if (propertyObject == null || propertyObject.Equals(null))
```
bddckr: Is this necessary to check for Unity's equality and thus checking whether it's destroyed or not?

Perhaps alternatively we can just cast propertyObject to UnityEngine.Object instead above and only need the first null check here then?

Editor/Data/Type/DefaultObservableListDrawer.cs
```diff
@@ -2,139 +2,30 @@
...
-        {
-            Type type = fieldInfo.FieldType;
-            while (type.BaseType != null)
-            {
-                type = type.BaseType;
-                if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(ObservableList<,>))
-                {
-                    return type.GenericTypeArguments[0];
-                }
+                SerializedObject referenceObject = new SerializedObject(propertyObject as Object);
```
bddckr: Use a direct cast instead. Or, if the above suggestion works, there is no need for a cast here in the first place.

PR #412, opened 2019-07-21
fix(Tracking): Teleport snap to floor incorrect

### LINKED ISSUE #411: Teleport snap to floor incorrect
Thanks @mattdavisgames . I dig further into the 'Comparer' used.
It is `SurfaceLocator.RayCastHitComparer` and its compare function is `return (int)(x.distance - y.distance);`
I think this is incorrect and should be `return x.distance.CompareTo(y.distance)`

Update:
confirmed that after changing the comparer function, the example scene house veranda/foundation can be reached correctly

_Originally posted by @fight4dream in https://github.com/ExtendRealityLtd/VRTK/issues/2042#issuecomment-509975640_

fight4dream:
Fixes #411

### DIFF
--- a/Runtime/Tracking/SurfaceLocator.cs
+++ b/Runtime/Tracking/SurfaceLocator.cs
@@ -28,7 +28,7 @@
             /// <inheritdoc />
             public virtual int Compare(RaycastHit x, RaycastHit y)
             {
-                return (int)(x.distance - y.distance);
+                return x.distance.CompareTo(y.distance);
             }
         }
 
--- a/Tests/Editor/Tracking/SurfaceLocatorTest.cs
+++ b/Tests/Editor/Tracking/SurfaceLocatorTest.cs
@@ -395,5 +395,52 @@ namespace Test.Zinnia.Tracking
             Object.DestroyImmediate(validSurface);
             Object.DestroyImmediate(searchOrigin);
         }
+
+        [UnityTest]
+        public IEnumerator NearestSurface()
+        {
+            Physics.autoSimulation = true;
+
+            GameObject validSurface = GameObject.CreatePrimitive(PrimitiveType.Cube);
+            GameObject validSurface2 = GameObject.CreatePrimitive(PrimitiveType.Cube);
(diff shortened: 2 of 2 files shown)


thestonefox:
The commit is fine but the commit message needs work.

Firstly the message needs to start with a lowercase:

`fix(Tracking): teleport snap to floor incorrect`

And there also needs to be a commit description explaining the reason for the commit. Just do a `git commit --amend` and update the message and force push.

fight4dream:
okay I amended it.

thestonefox:
The commit title is fine, but the commit description has a couple of issue:

The line length should not exceed 72 characters, it refers to things that are a part of the VRTK example (the veranda) which zinnia has nothing to do with.

Instead just have the description something like this:

```
The SurfaceLocator.RayCastHitComparer compare function was not
returning the various element distances in the correct sorted order
due to the comparison operation.

Using the `x.CompareTo(y)` operation ensures the element distance
order is correct when sorted.
```

Also, I think it's probably a good idea to include a Unit Test to prove the old way fails and the new way passes:

https://github.com/ExtendRealityLtd/Zinnia.Unity/blob/master/Tests/Editor/Tracking/SurfaceLocatorTest.cs

I'm not sure how easy it will be to replicate the issue in a test, but it would be great if it's possible so it's clear to see where the problem is and why this fixes it, otherwise this could just be viewed as a magic fix.

fight4dream:
I am trying to write a unit test. How should I show that the old way fails? Should I derive from `SurfaceLocator` and override its `bool FindAllCollisions(Ray tracerRaycast)` which uses the old `Comparer`?

thestonefox:
I think actually RaycastHitComparer should be moved out into its own class outside of the SurfaceLocator then the RaycastHitComparer should have a unit test testing its comparision.

But I dont think that work is part of this pull request. I think because the RaycastHitComparer is nested right now then it's pointless to do the unit test and try to ram it into the SurfaceLocator somewhere.

So with that, as long as the PR description is updated I think this is good to merge and I can raise a ticket about moving out the class in the future.

fight4dream:
done. added the relevant unit test as well

Issue #413, opened 2019-07-27
NUnit.Framework.Assert.AreEqual does not handle unity objects correctly


fight4dream:
* Version of Zinnia (master 990e8106c094bd5e31ee237a432a9a10bdb7dd6e)
* Version of the Unity software (Unity 2019.1.0f2)

### Steps to reproduce

1. go to SurfaceLocatorTest.cs
2. modify line 55 from `Assert.AreEqual(validSurface.transform, subject.surfaceData.Transform);` to `Assert.AreEqual(searchOrigin.transform, subject.surfaceData.Transform);`
3. go to unity and run the test SurfaceLocatorTest -> ValidSurface

### Expected behavior

The test fails because search origin is not the same as the valid surface returned from surface data

### Current behavior

The test succeeds.
I think the `Assert.AreEqual(object, object)` somehow used the `implicit operator bool` of Transform, so the result is the same as `Assert.AreEqual((bool)validSurface.transform, (bool)subject.surfaceData.transform)` which is not correct.

### Suggested fixes
1. Use `NUnit.Framework.Assert.AreSame(object, object)`. this will check by reference equal
2. Use `UnityEngine.Assertions.Assert.AreEqual<T>(T, T)`. this is specialized to use overloaded equality

### PS
notice that many unit tests are also using this incorrect `AreEqual` overload.

fight4dream:
Today I tried several approaches to fix this problem
1. `using Assert = UnityEngine.Assertions.Assert;` in every test files. Most assert calls are preserved, only a handful of them are missing such as `IsEmpty()` `GreaterOrEqual()` `LessOrEqual()` `NotNull()` which can be rewrite using `AreEqual()` `IsTrue()`. The only not so simple one missing is `Throws()`.
2. Change any `AreEqual(object, object)` calls that involves unity object to `AreSame`. There are a lot of places that uses `AreEqual(object, object)`, and NUnit does not provide overloading of int,float,string etc, which makes identifying unity object usage very difficult and not future fool-proof.
3. Inject, for each namespaces a derived class Assert : NUnit.Framework.Assert which contains some AreEqual overloads. It sort of works but every subnamespace has to inject the a derived class for the closest namespace class rule to look it up. This is not future fool-proof and prone to inconsistent error.

So, I would highly recommend to use approach 1.

### PS
I find from the correct `AreEqual` check that the test ObjectPointerTest -> EnterExitHover would fail.

PR #414, opened 2019-07-28
fix(Tests): use UnityEngine.Assertions.Assert in equality checking

fight4dream:
NUnit.Framework.Assert.AreEqual(object, object) uses the
implicit operator bool of UnityEngine.Object. So if two different
UnityEngine.Object are tested, AreEqual reports they are the same. #413


thestonefox:
The subject should explain what the commit does, not summarise the issue (but because this spans multiple scopes then the scope of `Tests` is fine)

`fix(Tests): use UnityEngine.Assertions.Assert in equality checking`

Issue #415, opened 2019-07-28
Incorrect test in ObjectPointerTest -> EnterExitHover


fight4dream:
* Version of Zinnia (master 990e8106c094bd5e31ee237a432a9a10bdb7dd6e)
* Version of the Unity software (Unity 2019.1.0f2)

the line 817 `Assert.AreEqual(blocker.transform, subject.HoverTarget.Transform);` 
should have been `Assert.AreEqual(blocker, subject.HoverTarget.CollisionData.transform.gameObject);`

#413

PR #416, opened 2019-07-28
fix(Pointers): check hover target collision data in select test

fight4dream:
The hover target should be the transform in collision data
Not the hover target transform (which is the object pointer)
#415


thestonefox:
This isn't the scope of `Tests` as it's specific to `Pointers`

Plus the subject should explain what this commit does:

`fix(Pointers): check hover target collision data in select test`

PR #417, opened 2019-08-04
feat(Data): remap a float value from a range to another

fight4dream:
The FloatRemapper component allows a float value to be remapped from
a range to another range, with an option to map the output linearly
or smooth step.

### DIFF
--- a/Runtime/Data/Type/Transformation/FloatRangeValueRemapper.cs
+++ b/Runtime/Data/Type/Transformation/FloatRangeValueRemapper.cs
@@ -1,0 +1,78 @@
+﻿namespace Zinnia.Data.Type.Transformation
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using System;
+    using Malimbe.PropertySerializationAttribute;
+    using Malimbe.XmlDocumentationAttribute;
+
+    /// <summary>
+    /// Transforms a <see cref="float"/> by remapping from a range to a new range.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


thestonefox:
I like this concept although I don't like the name, I don't think FloatRemapper conveys what's going on.

It's not actually working to remap a float but instead to remap a float between two ranges.

So rangeA -> 0 - 10
and rangeB -> 0 - 1

map 5 from rangeA to rangeB should output 0.5

I probably agree this isn't a conversion as you're not converting between types so leaving it in the Transformation namespace is fine.

But the name needs to reflect that it is remapping a float between float ranges.

Perhaps `FloatRangeValueRemapper`?

Also, in your class summary, you should include an example of how the thing works, see

https://github.com/ExtendRealityLtd/Zinnia.Unity/blob/master/Runtime/Data/Type/Transformation/Vector3Restrictor.cs

```
    /// <summary>
    /// Transforms a <see cref="Vector3"/> by clamping the coordinates to be within given boundaries.
    /// </summary>
    /// <example>
    /// input(5f,-5f,5f) -> bounds([-2f,2f], [-3f,3f], [-2f,2f]) = (2f,-3f,2f)
    /// input(5f,-5f,5f) -> bounds([-2f,2f], [1f,3f], [-2f,2f]) = (2f,1f,2f)
    /// </example>
```

fight4dream:
Thanks, I've changed accordingly.

PR #418, opened 2019-08-04
feat(Data): transform float to Vector3

fight4dream:
The FloatToVector3 component allows to transform a float array
to a Vector3.

### DIFF
--- a/Runtime/Data/Type/Transformation/Conversion/FloatToVector3.cs
+++ b/Runtime/Data/Type/Transformation/Conversion/FloatToVector3.cs
@@ -1,0 +1,78 @@
+﻿namespace Zinnia.Data.Type.Transformation.Conversion
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using System;
+    using Malimbe.XmlDocumentationAttribute;
+    using Malimbe.PropertySerializationAttribute;
+
+    /// <summary>
+    /// Transforms a float array to a Vector3.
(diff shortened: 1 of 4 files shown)


thestonefox:
I'm struggling to see how this is a transformation. It's basically constructing a vector3 from 3 individual float values.

Is there a use case for this feature?

fight4dream:
I am using it to construct a Vector3 from the output of float range value remapped and feed it into TransformScaleMutator. The mutator needs Vector3 input.
I see that there is FloatToVector2 so I copy it and make FloatToVector3.
I could provide an overload DoTransform(float value) that does all CurrentX/Y/Z + DoTransform. but FloatToVector2 doesn't have it so I keep it consistent.

It could be related to Grab scale action (I want to scale but not pinch scale logic). but I don't have an example in the ExampleScene. I am all ears for your advice.

image (1)

thestonefox:
I can see actually you've just copied the FloatToVector2 so you've done nothing untoward, but I wonder if Conversion is the right place for this stuff now I'm thinking about it

Like, you're not really converting a float to a vector2, you're building a vector2 on the fly from floats.

Like float to boolean or boolean to float make sense because you're converting

But then like Vector2ToFloat again i don't think is a conversion because you're extracting an element from a Vector2

Really there should be DataBuilders and DataExtractors along with DataConverters

But for the sake of this upheaval I think it's probably best to just leave this in the converter.

PR #420, opened 2019-08-16
fix(MomentProcessor): fix events to work correctly when a SRP is used

reznovVR:
This fixes the MomentProcessor to use correct events when a SRP is used

### DIFF
--- a/Runtime/Process/Moment/MomentProcessor.cs
+++ b/Runtime/Process/Moment/MomentProcessor.cs
@@ -5,6 +5,7 @@
     using Malimbe.XmlDocumentationAttribute;
     using Malimbe.PropertySerializationAttribute;
     using Zinnia.Process.Moment.Collection;
+    using UnityEngine.Rendering;
 
     /// <summary>
     /// Iterates through a given <see cref="MomentProcess"/> collection and executes the <see cref="IProcessable.Process"/> method on the given Unity game loop moment.
@@ -89,6 +90,11 @@
             }
         }
 
+        protected virtual void OnSrpCameraPreRender(ScriptableRenderContext context, Camera givenCamera)
+        {
+            Process();
+        }
+
         protected virtual void OnCameraPreRender(Camera givenCamera)
         {
             Process();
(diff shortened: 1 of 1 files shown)


thestonefox:
I think we leave this PR open as a way to help people fix it if they need it, but this fix alone in't enough to get the SRP stuff fully working so it's best to just stick to the official line of SRP isn't supported (yet)

Runtime/Process/Moment/MomentProcessor.cs
```diff
@@ -89,6 +90,11 @@ protected virtual void LateUpdate()
             }
         }
 
+        protected virtual void OnSRPCameraPreRender(ScriptableRenderContext context, Camera givenCamera)
```
bddckr:
```suggestion
        protected virtual void OnSrpCameraPreRender(ScriptableRenderContext context, Camera givenCamera)
```

I don't think it's worth typing it out, nor can I come up with a better name... But the general naming conventions forbid abbreviations from breaking the PascalCasing.

thestonefox:
The commits need a squash into one commit, plus the commit message needs updating

The commit message and description need to describe the purpose of the commit and the rationale behind it.

```
fix(MomentProcessor): allow PreRender to work with Unity SRP

The new Unity Scriptable Render Pipeline introduces a new way of
handling the event for when a camera begins rendering, meaning the
existing `Camera.onPreRender` event is no longer utilized. This
causes the `PreRender` moment to fail in the MomentProcessor because
the camera no longer has an `onPreRender` event when using the
scriptable render pipeline. This has been replaced by using the
`RenderPipelineManager` and the `beginCameraRendering` event which
can be used in place of the `onPreRender` if an SRP is being used.

This fix simply checks to see if the graphic settings are using a
scriptable render pipeline and uses the appropriate event for the
PreRender moment.
```

Note how the subject line is to the point on what the commit is about (also doesn't repeat the word fix as this is the commit type anyway) then the commit body goes into details about the reason why the commit was done.

PR #423, opened 2019-09-16
fix(Action): all/anyAction references destroyed actions

fight4dream:
If All/AnyAction references other actions that are destroyed,
it will generate null reference exception when all/any attempts to
remove listeners (when they are disabled or destroyed).

### DIFF
--- a/Runtime/Action/AllAction.cs
+++ b/Runtime/Action/AllAction.cs
@@ -63,7 +63,10 @@
 
             foreach (Action action in Actions.SubscribableElements)
             {
-                action.ActivationStateChanged.RemoveListener(OnActionActivationStateChanged);
+                if (action != null)
+                {
+                    action.ActivationStateChanged.RemoveListener(OnActionActivationStateChanged);
+                }
             }
         }
 
--- a/Runtime/Action/AnyAction.cs
+++ b/Runtime/Action/AnyAction.cs
@@ -63,7 +63,10 @@
 
             foreach (Action action in Actions.SubscribableElements)
             {
-                action.ActivationStateChanged.RemoveListener(OnActionActivationStateChanged);
+                if (action != null)
+                {
+                    action.ActivationStateChanged.RemoveListener(OnActionActivationStateChanged);
+                }
             }
         }
 
--- a/Tests/Editor/Action/AllActionTest.cs
+++ b/Tests/Editor/Action/AllActionTest.cs
@@ -245,6 +245,28 @@ namespace Test.Zinnia.Action
             Assert.IsFalse(deactivatedListenerMock.Received);
             Assert.IsFalse(changedListenerMock.Received);
         }
+
+        [Test]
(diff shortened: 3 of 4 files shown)


thestonefox:
can you add a test to prove this fixes things?

fight4dream:
Yes sure will make one on Wednesday

Runtime/Action/AllAction.cs
```diff
@@ -63,7 +63,10 @@ protected virtual void RemoveActionsListeners()
 
             foreach (Action action in Actions.SubscribableElements)
             {
-                action.ActivationStateChanged.RemoveListener(OnActionActivationStateChanged);
+                if (action)
```
bddckr: Code convention: We want to be explicit about what defines the "boolness" of objects, especially since [Unity overrides the equality of `UnityEngine.Object`s](https://blogs.unity3d.com/2014/05/16/custom-operator-should-we-keep-it/).
```suggestion
                if (action != null)
```

Runtime/Action/AnyAction.cs
```diff
@@ -63,7 +63,10 @@ protected virtual void RemoveActionsListeners()
 
             foreach (Action action in Actions.SubscribableElements)
             {
-                action.ActivationStateChanged.RemoveListener(OnActionActivationStateChanged);
+                if (action)
```
bddckr: Code convention: We want to be explicit about what defines the "boolness" of objects, especially since [Unity overrides the equality of `UnityEngine.Object`s](https://blogs.unity3d.com/2014/05/16/custom-operator-should-we-keep-it/).
```suggestion
                if (action != null)
```

Runtime/Action/AllAction.cs
fight4dream: should i make the actual change rather than just pressing the resolve conversation button?

bddckr: Doesn't matter how you apply the suggestions! All that matters in the end is that you squash the commits into one, leaving the nice subject and body for the message, explaining the change (and the "why").

fight4dream:
I have tried but it doesn't generate null reference exception during the test. May be it cannot mimic the OnApplicationQuit behaviour?

        [Test]
        public void RemoveActionListenersCorrectly()
        {
            subject.enabled = false;

            GameObject otherObject = new GameObject();
            MockAction actionA = otherObject.AddComponent<MockAction>();
            MockAction actionB = otherObject.AddComponent<MockAction>();

            actionA.SetIsActivated(false);
            actionB.SetIsActivated(false);

            subject.Actions.Add(actionA);
            subject.Actions.Add(actionB);

            subject.enabled = true;

            Object.DestroyImmediate(otherObject);

            subject.enabled = false;
        }

image

crazy. it is null-valued but its child is still accessible
but in the real application, it will generate null reference exception

thestonefox:
Looks good to me, not sure if @bddckr wants to give it a look over as he's marked as a reviewer as well.

PR #426, opened 2019-10-07
feat(Moment): treats a given MomentProcess collection as IProcessable.

fight4dream:
The CompositeProcess component allows to treat a
MomentProcessObservableList as an IProcessable.
This allows related moment processes to be organized
into a list and be nested in other list.

### DIFF
--- a/Runtime/Process/Moment/CompositeProcess.cs
+++ b/Runtime/Process/Moment/CompositeProcess.cs
@@ -1,0 +1,36 @@
+﻿namespace Zinnia.Process.Moment
+{
+    using Malimbe.PropertySerializationAttribute;
+    using Malimbe.XmlDocumentationAttribute;
+    using UnityEngine;
+    using Zinnia.Process.Moment.Collection;
+
+    /// <summary>
+    /// Processes a list of <see cref="MomentProcess"/> as if it's a single <see cref="IProcessable"/> process.
+    /// </summary>
+    public class CompositeProcess : MonoBehaviour, IProcessable
+    {
+        /// <summary>
+        /// A collection of <see cref="MomentProcess"/> to process.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public MomentProcessObservableList Processes { get; set; }
+
+        /// <summary>
+        /// Iterates through <see cref="Processes"/> and calls <see cref="MomentProcess.Process"/> on each one.
+        /// </summary>
+        public void Process()
+        {
+            if (Processes == null)
+            {
+                return;
+            }
+
+            foreach (MomentProcess currentProcess in Processes.NonSubscribableElements)
+            {
+                currentProcess.Process();
+            }
+        }
+    }
+}
--- a/Runtime/Process/Moment/CompositeProcess.cs.meta
+++ b/Runtime/Process/Moment/CompositeProcess.cs.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: 378ecdf468e07574db371a118cd2c5cd
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Tests/Editor/Process/Moment/CompositeProcessTest.cs
+++ b/Tests/Editor/Process/Moment/CompositeProcessTest.cs
@@ -1,0 +1,126 @@
+﻿using Zinnia.Process;
+using Zinnia.Process.Moment;
+using Zinnia.Process.Moment.Collection;
+
+namespace Test.Zinnia.Process.Moment
+{
+    using NUnit.Framework;
+    using UnityEngine;
+    using Assert = UnityEngine.Assertions.Assert;
+
+    public class CompositeProcessTest
+    {
+        private GameObject containingObject;
+        private CompositeProcess subject;
+
+        [SetUp]
+        public void SetUp()
+        {
+            containingObject = new GameObject();
+
+            subject = containingObject.AddComponent<CompositeProcess>();
+            subject.Processes = containingObject.AddComponent<MomentProcessObservableList>();
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            Object.DestroyImmediate(containingObject);
+        }
+
+        [Test]
+        public void Process()
+        {
+            MockProcessable mockProcessable1 = new MockProcessable();
+            MomentProcess momentProcess1 = containingObject.AddComponent<MomentProcess>();
+            momentProcess1.Source = new ProcessContainer
+            {
+                Interface = mockProcessable1
+            };
+            MockProcessable mockProcessable2 = new MockProcessable();
+            MomentProcess momentProcess2 = containingObject.AddComponent<MomentProcess>();
+            momentProcess2.Source = new ProcessContainer
+            {
+                Interface = mockProcessable2
+            };
+            subject.Processes.Add(momentProcess1);
+            subject.Processes.Add(momentProcess2);
+
+            Assert.IsFalse(mockProcessable1.WasProcessCalled);
+            Assert.IsFalse(mockProcessable2.WasProcessCalled);
+            subject.Process();
+            Assert.IsTrue(mockProcessable1.WasProcessCalled);
+            Assert.IsTrue(mockProcessable2.WasProcessCalled);
+        }
+
+        [Test]
+        public void ProcessNestedList()
+        {
+            MockProcessable mockProcessable1 = new MockProcessable();
+            MomentProcess momentProcess1 = containingObject.AddComponent<MomentProcess>();
+            momentProcess1.Source = new ProcessContainer
+            {
+                Interface = mockProcessable1
+            };
+            MockProcessable mockProcessable2 = new MockProcessable();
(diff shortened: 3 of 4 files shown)


thestonefox:
This is a step in the right direction, but I wonder if it a touch misleading? It surely will order the game loop moments (Update, FixedUpdate, LateUpdate) but any camera event (PreCull, PreRender) will still be in whatever order the MomentProcessor became enabled.

Perhaps OnEnable of this script it could go through all of the MomentProcessors in the list and disable them (to force the unregister to occur) then enable them in order?

It would be a somewhat costly loop as you'd need to loop twice, but only does it in on enable.

@bddckr any thoughts on that?

bddckr:
Few random thoughts as I'm not sure what this is solving and what you're onto there @thestonefox:

* Why do we need this PR when you can just add those processes you'd add to the new ObsList field to the processor's list instead?
* If the idea is to allow easier to ship composable prefabs etc.: This doesn't tackle the main issue, which is wiring up to *shared processors*.
* I don't know why your points are of concern for the changes raised in this PR @thestonefox as I don't see this PR touching the issue of processors at all. I.e. there's no confusion as this changeset "just" allows to provide lists of processes in a different way. It does *not* tackle any wiring to a processor.

I think this PR feels like a good step into the direction of "composable prefabs that have processes, thus should come with processors to be self-contained, yet we want to allow explicit control and a bit of a performance optimization by _easily_ wiring up a shared processor to be used instead", but it's still missing that crucial step of the actual processor change.

Is your idea to leave that up to the user and ship prefabs with a CompositeProcess, but no processors perhaps @fight4dream?

thestonefox:
I believe the issue being addressed was if you have multiple moment processors in a scene then you can't control the order in which each of the processes occurs.

So yeah in short it's how you deal with independent moment processors in the scene but still control the absolute order of them.

I don't think this PR solves that problem though

fight4dream:
There may be other issues/PR in you guys mind and the design direction of moment processor/process, may be I don't have a big picture of it so I do not understand the involvement of a MomentProcessor change.

Let me go through my thoughts and how do I come up with CompositeProcess in my project first.
0. I searched and cannot find any derived class from MomentProcess. So I am confident that for any 'loopable' thing I should be using only IProcessable.
0a. as to why a MomentProcess is itself a IProcessable, I have no idea why and how it is used that way.
1. To actually execute the MomentProcess (processables), we have MomentProcessor. Same with the standard Unity functions, different moments are occurring in their specific time following https://docs.unity3d.com/Manual/ExecutionOrder.html
1a. So I have no problem if I happen to use 2 moment processors with different moments. because just like Update() and FixedUpdate(), if my logic depends on their order I should be screwed up anyway.
2. Different MomentProcessors with the same moment (e.g. Update), just like various scripts using Update loop, their execution order is undetermined. 
2a. However, the processables in a single MomentProcessor is in the listed order. this is useful.
3. For my project, I often have to use a list of alignment processables for equipments. e.g. my Bow when 1-handed, will only align transform with the grabbing hand. when other hand grab the string for arrow, these happens in this order (1) align bow position with grabbing hand (2) align bow rotation to look at other hand (3) align arrow position with other hand (4) align arrow rotation to look at grabbing hand (5) align string position with other hand
3a. while this bow alignment is self contained, if it is incorporated with the player's ik model and gloves, the different MomentProcessors (with the same moment) will have a race condition due to above point 2. and for some reasons, I want to have ik model, gloves, bow alignments in that order in particular.
4. it is thus natural to follow that "a list of processables" should be eligible to be put into and orderable in MomentProcessor's list.

solutions thought through:
0. MomentProcessObservableList can be IProcessable. this is a major component in vrtk, so I somehow have a gut feeling I should not touch this casually. (in the retrospection, this could have been my PR and I would change MomentProcessor.Processes back to a single ProcessContainer Process. but this is a breaking change. so a no no.)
1. write a script to add all processables from source lists onto target list. there are some concern. because my game would instantiate and destroy equipments at runtime. I feel that maintaining the add/remove a bunch of processables would be problematic.
2. thus I come up with a CompositeProcess that wraps a list into IProcessable. Because i already have a script that is responsible to instantiate/remove equipment, it will do the find/add/remove CompositeProcess into the centralized processor. so I do not touch any rewiring in this PR. as I think we don't possibly know how our user will order the composite lists.

bddckr:
> 1. I searched and cannot find any derived class from MomentProcess.

Because you just implement `IProcessable`. No need to subclass a base class - it's just an interface.

>    1a. as to why a MomentProcess is itself a IProcessable, I have no idea why and how it is used that way.

`MomentProcess` is just a helper to get to define `IProcessable`s in Unity (which doesn't like interfaces in the editor).

> 1. MomentProcessObservableList can be IProcessable. this is a major component in vrtk, so I somehow have a gut feeling I should not touch this casually. (in the retrospection, this could have been my PR, then a MomentProcess can hold a list already. then it is up to us to also change MomentProcessor.Processes back to a single ProcessContainer Process. but this is a breaking change. so a no no.)

`MomentProcessObservableList` is literally just an `ObservableList`, "fixing" the generic type argument (to make Unity's serializer and thus the editor-integration happy). An observable list just holds objects. It doesn't do anything. Thus it should never be `IProcessable`.

> 2. write a script to add all processables from source lists onto target list. there are some concern. because my game would instantiate and destroy equipments at runtime. I feel that maintaining the add/remove a bunch of processables would be problematic.
> 3. thus I come up with a CompositeProcess that wraps a list into IProcessable. Because i already have a script that is responsible to instantiate/remove equipment, it will do the find/add/remove CompositeProcess into the centralized processor. so I do not touch any rewiring in this PR. as I think we don't possibly know how our user will order the composite lists.

This is where I'm getting confused. Your issue boils down to "I want to merge lists to be used in a single processor" if I understood it correctly.

Why don't you just add your objects all (in your defined order) to a single processor's list? Why do you have to introduce a list of processes in between, just to then again manage the same addition/removal logic when adding to your one processor?

Also: You're saying it's a tricky thing to get the add/remove right, yet you then continue to explain that you do exactly that, but instead of adding N object's M processes you now add N object's single `CompositeProcess`es? To me you're solving the same thing, so might as well not introduce a new type to sit in between and just use what's already available:
A processor takes a list of processes to enumerate and call `Process` on. Just add/remove to/from that one's list directly.

fight4dream:
“MomentProcess is just a helper to get to define IProcessables in Unity”
This is true. And you defined intervals, active to process or not. 

“(which doesn't like interfaces in the editor).”
You imply moment process has to implement IProcessable in order to do this, I disagree on this because ProcessContainer is the class you use to show a IProcessable to inspector. MomentProcess being a IProcessable is not related to this

fight4dream:
> You're saying it's a tricky thing to get the add/remove right, yet you then continue to explain that you do exactly that, but instead of adding N object's M processes you now add N object's single CompositeProcesses?

agree. I think i could do it carefully by code it will be fine.

I want to tackle this decision in another angle,
consider the follow production process:

Gloves (don't mind the CompositeProcess, pretend they're MomentProcessors)
image

Avatar which the artists then decided to embed the gloves in it
image
with composite - drag one process from gloves onto the avatar list
without composite - disable gloves moment processor, drag 4 things from gloves onto avatar list

PlayerExample, embed with one variant of the Avatar
image
with composite - drag one process from avatar onto player list (don't have to care about gloves)
without composite - disable avatar moment processor, drag 4 things from avatar, AND drag 4 things from gloves

Would this justify the use of a 'list as IProcessable' ?

bddckr:
Yeah I see how that is useful. So you're basically just aggregating them to make manual composition "further up", without tackling the issue of that composition at all.

That is fine. I see the usefulness in this PR. Reviewing in a bit

thestonefox:
It still doesn't handle the composition of camera events though. Which could lead to confusion.

bddckr:
I still don't get what you're onto there @thestonefox. This is a process, not a processor. Only the **processors** have such moments, thus this PR doesn't change anything in that regard!

thestonefox:
Maybe I need to read what's actually going on then :)

fight4dream:
What the camera events issue that you want to solve? Could you reference the issue number so i can understand better?

If i understand correctly, is it about, say, two different moment processors having the same moment then their order is undetermined? It is indeed undetermined as if those Update() calls are undetermined as warned by Unity themselves. 

In case there is some setup in the farm scene that you need multiple of them in specific order, actually the composite process can help you and our users to follow them more easily. But i dont intend the wiring or rewiring be not automatic. Like i said, we cant know what the order should be

thestonefox:
> What the camera events issue that you want to solve? Could you reference the issue number so i can understand better?
> 
> If i understand correctly, is it about, say, two different moment processors having the same moment then their order is undetermined? It is indeed undetermined as if those Update() calls are undetermined as warned by Unity themselves.
> 
> In case there is some setup in the farm scene that you need multiple of them in specific order, actually the composite process can help you and our users to follow them more easily. But i dont intend the wiring or rewiring be automatic. Like i said, we cant know what the order should be

Ignore me, @bddckr is more sensible than I am and actually reads commits :)

thestonefox:
Squash dem commits though ;)

fight4dream:
Sadly not besides my computer atm. I think i cant squash on mobile

fight4dream:
What will be the correct commands to squash it?

thestonefox:
> What will be the correct commands to squash it?

do a fixup in an interactive git rebase

so like

`git rebase -i HEAD~10`

look for all of your commits and change everything after your initial commit to a letter of `f`

extendreality:
`git commit --amend`

change your git commit message so the sentence starts lowercase

`feat(Moment): processes a list of MomentProcess as single IProcessable.`

PR #434, opened 2019-10-12
feat(Attribute): provide custom inspector text for inspector properties - resolves #431

### LINKED ISSUE #431: Attribute to allow changing the inspector text for a parameter
Unity by default will convert a field/property based on its casing into separated words uppercasing the first letter:

`myParamIsThis` becomes `My Param Is This`

But if there is something that is an acronym like

`mySqlQuery` then it becomes `My Sql Query` which really should read `My SQL Query`

Here is a potential solution for such an attribute:

https://answers.unity.com/questions/1487864/change-a-variable-name-only-on-the-inspector.html

extendreality:
The new CustomInspectorText attribute allows an inspector field to have
custom text instead of Unity using the field/property name for the
inspector text.

This only works with simple fields at the moment and does not work with
nested fields such as the ObservableList Drawer.

### DIFF
--- a/Editor/Data/Attribute/CustomInspectorTextAttributeDrawer.cs
+++ b/Editor/Data/Attribute/CustomInspectorTextAttributeDrawer.cs
@@ -1,0 +1,14 @@
+﻿namespace Zinnia.Data.Attribute
+{
+    using UnityEngine;
+    using UnityEditor;
+
+    [CustomPropertyDrawer(typeof(CustomInspectorTextAttribute))]
+    public class CustomInspectorTextAttributeDrawer : PropertyDrawer
+    {
+        public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
+        {
+            EditorGUI.PropertyField(position, property, new GUIContent(((CustomInspectorTextAttribute)attribute).customText));
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


Editor/Data/Attribute/CustomInspectorTextAttributeDrawer.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    using UnityEngine;
+    using UnityEditor;
+
+    [CustomPropertyDrawer(typeof(CustomInspectorTextAttribute))]
+    public class CustomInspectorTextAttributeDrawer : PropertyDrawer
+    {
+        public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
+        {
+            EditorGUI.PropertyField(position, property, new GUIContent((attribute as CustomInspectorTextAttribute).customText));
```
bddckr:
```suggestion
            EditorGUI.PropertyField(position, property, new GUIContent(((PropertyDrawer)attribute).customText));
```

We know it's that type, due to the `CustomPropertyDrawer` attribute telling Unity that.

bddckr:
While I just reviewed the technical details about this I'm still wondering what the use for this is.

Changing how API is displayed in the inspector means users will have a harder time discovering the actual API to reference by code.

extendreality:
The only time I think we'd use it is if there was like acronyms that we wanted to uppercase.

Editor/Data/Attribute/CustomInspectorTextAttributeDrawer.cs
extendreality: image

this change throws an error

bddckr: Ha damn, that was a copy and past error on my behalf! 😄 I edited the suggestion (because GitHub doesn't seem to allow suggestions in comments down the comment thread chain).

PR #435, opened 2019-10-12
fix(Association): perform null check before accessing associations

extendreality:
An exception error would occur if the Association list was set up
but contained empty elements, which is the default state for Unity.

This has been fixed by simply doing null checks before accessing data.

### DIFF
--- a/Runtime/Association/GameObjectsAssociationActivator.cs
+++ b/Runtime/Association/GameObjectsAssociationActivator.cs
@@ -29,9 +29,19 @@
         /// </summary>
         public virtual void Activate()
         {
+            if (Associations == null)
+            {
+                return;
+            }
+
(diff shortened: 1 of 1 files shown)


Runtime/Association/GameObjectsAssociationActivator.cs
```diff
@@ -83,6 +93,11 @@ public void Process()
...
         protected virtual void Awake()
         {
+            if (Associations == null)
+            {
+                return;
+            }
+
             foreach (GameObjectsAssociation association in Associations.NonSubscribableElements)
             {
                 foreach (GameObject associatedObject in association.GameObjects.NonSubscribableElements)
```
bddckr: This seems to be missing the applied change to use
```c#
                if (association == null)
                {
                    continue;
                }
```

PR #436, opened 2019-10-13
UPM package & CI+CD - resolves #430

### LINKED ISSUE #430: Make compatible with the Unity Package Manager and remove hard dependencies
At the moment, Zinnia is included with other projects as a sub domain, or hard copied into the project.

Adding a `package.json` to the root of the repo with the relevant contents will allow it to become a Unity Package.

This also means that the dependency of Malimbe should be hard removed from Zinnia and referenced as a dependency in the `package.json`

To ensure the package is available to Unity projects via UPM it will also need to be pushed to a registry like npmjs

This should all be automated via a CI+CD pipeline (as like with Malimbe).

bddckr:


bddckr:
Once merged this will be kicking off releases whenever we push to `master`.

The first version will be 1.0.0. We can change that but I figured we might as well treat this first release as the stable 1.0.0 of Zinnia from here on!

SimonDarksideJ:
Nice, great to see VRTK moving over to UPM and devops!  Well done.

bddckr:
>Although you might want to do a seed package asset to register the scoped registry automatically for those unfamiliar with editing their project manifest.

I don't really like that and instead would like to leave it up to Unity to improve their tooling. At the end of the day the documentation on how to do it is given by us and Unity, so I think we'll be fine.

PR #437, opened 2019-10-13
feat(Velocity): multiply velocity by vector rather than float

thestonefox:
*Breaking Change*

The VelocityMultiplier only allowed for a float to be multiplied
to the whole velocity vector making the multiplication very linear
and not able to affect different parts of the velocity by different
values.

This has now been changed so the multiplier is actually a Vector3,
meaning it can affect different elements with different multipliers.

This is also useful if a certain velocity axis wants to be removed
by simply multiplying that axis by 0.

### DIFF
--- a/Runtime/Tracking/Velocity/VelocityMultiplier.cs
+++ b/Runtime/Tracking/Velocity/VelocityMultiplier.cs
@@ -21,13 +21,13 @@
         /// </summary>
         [Serialized]
         [field: DocumentedByXml]
-        public float VelocityMultiplierFactor { get; set; } = 1f;
+        public Vector3 VelocityMultiplierFactor { get; set; } = Vector3.one;
         /// <summary>
         /// The amount to multiply the source angular velocity by.
         /// </summary>
         [Serialized]
         [field: DocumentedByXml]
-        public float AngularVelocityMultiplierFactor { get; set; } = 1f;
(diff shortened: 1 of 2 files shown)


Runtime/Tracking/Velocity/VelocityMultiplier.cs
```diff
@@ -21,13 +21,22 @@ public class VelocityMultiplier : VelocityTracker
...
         /// <summary>
         /// The amount to multiply the source angular velocity by.
         /// </summary>
         [Serialized]
         [field: DocumentedByXml]
-        public float AngularVelocityMultiplierFactor { get; set; } = 1f;
+        public Vector3 AngularVelocityMultiplierFactor { get; set; } = Vector3.one;
+
+        /// <summary>
+        /// The container for muliplied velocity.
```
bddckr: Why is this caching of the results necessary? Can we remove these two `protected` fields?

thestonefox: Just so it's not doing a `new Vector3` in place

because the source is a return value so you can't do

`source.GetVelocity().Scale(vec)`

because source.GetVelocity is a calculation returned from another method

bddckr: These are value types, the copying is happening anyway. Use a local variable in the methods instead, e.g. `Vector3 multipliedVelocity = Source.GetVelocity();`.

PR #444, opened 2019-10-23
feat(Haptics): create haptic pattern based on AudioSource

fight4dream:
This haptic generated reflects the current playing audio even if
the audio is looping, delayed, volume/pitch modulated.

### DIFF
--- a/Runtime/Haptics/AudioClipHapticPulser.cs
+++ b/Runtime/Haptics/AudioClipHapticPulser.cs
@@ -2,45 +2,25 @@
 {
     using UnityEngine;
     using System.Collections;
-    using Malimbe.XmlDocumentationAttribute;
     using Malimbe.PropertySerializationAttribute;
+    using Malimbe.XmlDocumentationAttribute;
 
     /// <summary>
     /// Creates a haptic pattern based on the waveform of an <see cref="UnityEngine.AudioClip"/> and utilizes a <see cref="Haptics.HapticPulser"/> to create the effect.
     /// </summary>
-    public class AudioClipHapticPulser : HapticProcess
+    public class AudioClipHapticPulser : RoutineHapticPulser
     {
-        /// <summary>
-        /// The pulse process to utilize.
-        /// </summary>
-        [Serialized]
-        [field: DocumentedByXml]
-        public HapticPulser HapticPulser { get; set; }
         /// <summary>
         /// The waveform to represent the haptic pattern.
         /// </summary>
         [Serialized]
         [field: DocumentedByXml]
         public AudioClip AudioClip { get; set; }
-        /// <summary>
-        /// Multiplies the current audio peak to affect the wave intensity.
-        /// </summary>
-        [Serialized]
-        [field: DocumentedByXml]
-        public float IntensityMultiplier { get; set; } = 1f;
 
         /// <summary>
         /// The size of the audio buffer.
         /// </summary>
         protected const int BufferSize = 8192;
-        /// <summary>
-        /// A reference to the started routine.
-        /// </summary>
-        protected Coroutine hapticRoutine;
-        /// <summary>
-        /// The original intensity of <see cref="HapticPulser"/> to reset back to after the process is complete.
-        /// </summary>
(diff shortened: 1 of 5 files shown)


Runtime/Haptics/AudioSourceHapticPulser.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            StopCoroutine(hapticRoutine);
+            hapticRoutine = null;
+            HapticPulser.Cancel();
+            ResetIntensity();
+        }
+
+        /// <summary>
+        /// Resets the <see cref="Haptics.HapticPulser.Intensity"/> back to it's original value.
```
bddckr:
```suggestion
        /// Resets the <see cref="Haptics.HapticPulser.Intensity"/> back to its original value.
```

bddckr:
Also: I _looove_ this idea. 💓

fight4dream:
base class name `AudioHapticPulser` ?

fight4dream:
Oh i guess i overlooked it. 
Btw, any suggestions on the commit title/message?

bddckr:
Looking good to me, but perhaps should mention the refactoring changes:
```
feat(Haptics): create haptic pattern based on AudioSource

This haptic generated reflects the current playing audio even if the
audio is looping, delayed, volume/pitch modulated. Since this new
implementation shares the same core logic as the existing
AudioClipHapticPulser the shared logic has been extracted into an
abstract superclass.
```

Runtime/Haptics/AudioClipHapticPulser.cs
```diff
@@ -1,91 +1,42 @@
 ﻿namespace Zinnia.Haptics
 {
-    using UnityEngine;
     using System.Collections;
-    using Malimbe.XmlDocumentationAttribute;
+    using UnityEngine;
```
thestonefox: As before, we put `Unity*` before `System*` (for now)

Runtime/Haptics/AudioSourceHapticPulser.cs
```diff
@@ -0,0 +1,58 @@
+﻿namespace Zinnia.Haptics
+{
+    using System.Collections;
```
thestonefox: apply consistent namespace order again

Runtime/Haptics/RoutineHapticPulser.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace Zinnia.Haptics
+{
+    using System.Collections;
```
thestonefox: and here

thestonefox:
I've tested this in Unity and I only get a single buzz for the first frame, it's not following the audioclip pattern.

I would assume this set up:
image

Would give me the same output as this set up:
image

fight4dream:
The usage for audio source haptics is 
1. AudioSource.Play() or other methods to make it playing
2. Then call haptics.Begin()

The first buzz is actually a bug, if the audio source isnt playing, it shouldnt buzz at all and should immediately end the routine

thestonefox:
Actually i figured out what i was doing wrong. I got it to work, but then after the loop I got 

```
IndexOutOfRangeException: Index was outside the bounds of the array.
Zinnia.Haptics.AudioSourceHapticPulser+<HapticProcessRoutine>d__7.MoveNext () (at Assets/Zinnia.Unity/Runtime/Haptics/AudioSourceHapticPulser.cs:50)
UnityEngine.SetupCoroutine.InvokeMoveNext (System.Collections.IEnumerator enumerator, System.IntPtr returnValueAddress) (at C:/buildslave/unity/build/Runtime/Export/Scripting/Coroutines.cs:17)
```

fight4dream:
Ok i will have it fix soon. Any comment on its usage for now?
I wish there is a Played event from audio source but nope

fight4dream:
i've changed it to use OnAudioFilterRead()
hence, this audio source haptic pulser need to be placed after audio source as well as other audio filters

thestonefox:
Not sure if @bddckr wants to have a final review. If not, then I'm happy to merge it

Issue #445, opened 2019-10-23
IProcessable that wraps up events


fight4dream:
I come up with a wrapper when Process() it will invoke events.
![image](https://user-images.githubusercontent.com/22952143/67419624-a24e0b00-f5ff-11e9-98e8-c00f39a2acd6.png)

this one is for no arguments, probably naming should be EmptyEventProxyProcessable
would it be a candidate to add to zinnia? as well as others such as FloatEventProxyProcessable

thestonefox:
so what does this do? when a momentprocess has finished processing it emits an event to say the moment has done processing?

thestonefox:
or is this something that will call an event every loop of the given moment?

fight4dream:
the later. to call an event every loop of the given moment

thestonefox:
i think that's pretty useful. I think that may actually solve a problem with the teleport/surface locator issue.

So you can basically have something at the end of a MomentProcessor which is an event that you want to do something.

You can then do kinda like "wait till end of frame" type stuff with those events.

I don't think this is a Proxy though

it's more like `EmptyEventProcess : EventProcessable`

@bddckr any thoughts on this?

bddckr:
So you basically get to call methods from the inspector as part of a specific moment?

Sure I like it. This basically means the process abstraction goes all the way and allows being explicit about a lot of things (as they can all be nested and chained together), right there without resorting to code.

In terms of implementation details: I'd expect to have the following classes:
* `abstract` base class `EventProcess` (or remove if the next two don't share anything)
* `abstract` intermediate base class `EventProcess<TValue>`
* specific implementation `EmptyEventProcessable : EventProcess` for calling empty methods

---

A few things that reduce the usefulness though:
1. No support for calling any method without having a specific `EventProcessable<TValue>` subclass created in advance, that matches the signature of the method to call.
2. Passing on multiple arguments to a method (i.e. calling any method that takes more than one argument) means one can't even just inherit from the `TValue` one as one needs _multiple_ values.
3. Arguments passed can only be constant. Unity doesn't support a lot of types here at all, even a `Vector3` can't be used.

**My current opinion, due to these, is to only offer the `EmptyEventProcess` class.** Instead of creating a specific subclass for all the calling needs us and users would be better off just writing the method call in code already. Additionally the in-code solution doesn't have issue 3 - one can retrieve and propagate values as needed. Unity events aren't a replacement for visual scripting after all.

[...]

PR #449, opened 2019-10-24
feat(Process): a IProcessable that emits event when processed.

fight4dream:
Allows to call any no argument methods for a given moment.

### DIFF
--- a/Runtime/Process/EventProcess.cs
+++ b/Runtime/Process/EventProcess.cs
@@ -1,0 +1,26 @@
+﻿namespace Zinnia.Process
+{
+    using UnityEngine;
+    using UnityEngine.Events;
+    using Malimbe.XmlDocumentationAttribute;
+
+    /// <summary>
+    /// Raises a <see cref="UnityEvent"/> when processed.
+    /// </summary>
+    public class EventProcess : MonoBehaviour, IProcessable
+    {
+        /// <summary>
+        /// Emitted when processed.
+        /// </summary>
+        [DocumentedByXml]
+        public UnityEvent Event = new UnityEvent();
+
+        /// <summary>
+        /// Emits <see cref="Event"/>.
+        /// </summary>
+        public void Process()
+        {
+            Event?.Invoke();
+        }
+    }
+}
--- a/Runtime/Process/EventProcess.cs.meta
+++ b/Runtime/Process/EventProcess.cs.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: adf4ef1e9383a1d469d3c5c5eebcb6b0
+MonoImporter:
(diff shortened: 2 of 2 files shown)


fight4dream:
PR for #445

Runtime/Process/EventProcess.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Zinnia.Process
+{
+    using Malimbe.XmlDocumentationAttribute;
+    using UnityEngine;
+    using UnityEngine.Events;
+
+    /// <summary>
+    /// Emit <see cref="UnityEvent"/> when processed.
```
bddckr:
```suggestion
    /// Raises a <see cref="UnityEvent"/> when processed.
```

"Raising an event" is proper C# terminology. I know the existing codebase talks about "emitting" most of the time, but we should start getting this right in new changes at least.

Runtime/Process/EventProcess.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Zinnia.Process
+{
+    using Malimbe.XmlDocumentationAttribute;
```
thestonefox: The Malimbe namespace should appear below the UnityEngine namespace

```
using UnityEngine;
using UnityEngine.Events;
```

fight4dream: What is the namespace sorting convention?

I am using the vs default which  is System first, then any other by alphabetical order. 

I would like to know your specific ordering

thestonefox: Right now it's a dumb one:

```
UnityEngine.*
System.*
Malimbe.*
Zinnia.*
VRTK.*
```

We're going to move to sorted alphabetical at some point, but we're trying to keep things consistent till then

thestonefox: I think our plan is to actually do something like this

```
System.*
Unity.*
Malimbe.*
Zinnia.*
VRTK.*
```

Starting with the most base namespace first, working down the layers.

fight4dream: Consider your future plan of adding other namespaces such as this:
image

It seems logical to extend your scheme by traversing in depth first ascending order then put the list reversed. 
```
System
UnityEngine
Malimbe
Zinnia
OrangeJuice
AppleJuice
VRTK
```
I think this could be a bit hard for maintenance. 
 i would prefer alphabetical order (may with the exception of System depends on which vs settings we adopt to), so that we can just “format document” and let the IDE do it for us

thestonefox: I think you can define the namespaces in the IDE to group them in the logical order you want.

I'm not overly fussed if it's all alphabetical. As long as it's consistent.

PR #459, opened 2019-10-28
Make CompositeProcess derive from MomentProcess directly

dustinlacewell:
Currently `CompositeProcess` instances cannot be assigned to `MomentProcessObservableList` instances. This change makes `CompositeProcess` derive from `MomentProcess` directly, which allows interoperability with `MomentProcessObservableList` and seems to conveniently work as expected without any breakage from what I could tell.


thestonefox:
This isn't how you use IProcessables. You apply them to a MomentProcess wrapper

Issue #460, opened 2019-10-29
AudioSourceHapticPulser can't be placed on another game object


fight4dream:
the audio source haptic pulser uses filter read, it has to be right beneath the audio source.
but the way it is presented in the inpsector deceived us to think it can be placed elsewhere.

![image](https://user-images.githubusercontent.com/22952143/67784442-c3eb3e80-faa6-11e9-9c57-fae7d74a627b.png)

### Solution

Add a hidden component to filter read at runtime, just like CollisionTracker will add CollisionTrackerDisabledObserver

thestonefox:
https://docs.unity3d.com/ScriptReference/RequireComponent.html

this adds it at edit time

fight4dream:
it is slightly different. i want the haptic pulser free to be on any object, not necessarily stick to audio source

PR #461, opened 2019-10-29
fix(Haptics): allow AudioSourceHapticPulser be placed on other object

fight4dream:
The pulser uses OnAudioFilterRead, it has to be placed right beneath the audio source.
But the way it is presented in the inpsector deceived us to think it can be placed elsewhere.
This fix allows it to be placed on other object.
Also fixed not getting pulse due to frame lag behind audio thread too much.


fight4dream:
#460

fight4dream:
i've used this implicit version in a project and discovered that, if i would like to have both left/right hand haptics on the same audio source, they each will add their copy of observer component to the audio source. whereas the explicit version, we only need one copy of observer, and it can be shared to both hand haptics.

Issue #467, opened 2019-11-04
Want a way for Vector3 to magnitude 


fight4dream:
### Situation

my current project requires detecting speed over threshold a lot.
The data flow is
CollisionTracker enter -> CollisionNotifierProxyEmitter emit (Check ColliderSource valid rule) -> VelocityEmitter emit -> [Vector3 to float (by magnitude)] -> FloatRangeValueRemapper (map to 0,1) -> FloatToBoolean -> Activated/Deactivated
The missing link is Vector3 to float (by magnitude).
As a reference, Vector2ToFloat is extracting components (x or y).

### Suggestion

1. Add Vector3ToFloat, it can extract (by x, y, z or magnitude)
2. Retro add extract (by magnitude) to Vector2ToFloat
3. (optional) ObjectDistanceComparator can be refactored to not have the threshold functionality, rather the data flow can be
ObjectDistanceComparator Process -> ObjectDistanceComparatorEventDataExtractor .DistanceExtracted -> FloatRangeValueRemapper (map to 0,1) -> FloatToBoolean -> Activated/Deactivated

### Alternative suggestion

1. Retro add VelocityEmitter EmitSpeed() and EmitAngularSpeed() and the corresponding float events (just like the ObjectDistanceComparator does emit both a Vector3 and its magnitude)
then the data flow is
CollisionTracker enter -> CollisionNotifierProxyEmitter emit (Check ColliderSource valid rule) -> VelocityEmitter emit speed -> FloatRangeValueRemapper (map to 0,1) -> FloatToBoolean -> Activated/Deactivated

thestonefox:
I'm ok with the `Vector3ToFloat` with additional extraction floats (and retro added to Vector2ToFloat as well).

But i'd add for all the properties:

* magnitude
* normalized
* sqrMagnitude

The property name `Coordinate` doesn't mean all that much then, it should be `Property To Extract` but this would be a breaking change and break a bunch of things so i'm not fussed if it's left as `Coordinate to Extract`

thestonefox:
Also, I don't mind the `VelocityEmitter` emitting the speed as well as velocity.

fight4dream:
VelocityEmitter is in Tracking namespace,
Vector2/3ToFloat is in Data namespace,
feat(Tracking) or feat(Data) shall be used for the PR commit title?

thestonefox:
The can be in 2 separate commits, therefore have 2 different commit titles

[...]

PR #470, opened 2019-11-06
feat(Data): add properties to extract float from Vector2 and Vector3

fight4dream:
#467

### DIFF
--- a/Runtime/Tracking/Velocity/VelocityEmitter.cs
+++ b/Runtime/Tracking/Velocity/VelocityEmitter.cs
@@ -17,12 +17,20 @@
         /// Defines the event with the <see cref="Vector3"/>.
         /// </summary>
         [Serializable]
-        public class UnityEvent : UnityEvent<Vector3>
+        public class Vector3UnityEvent : UnityEvent<Vector3>
         {
         }
 
         /// <summary>
-        /// The source <see cref="VelocityTracker "/> to receive the velocity data from.
+        /// Defines the event with the <see cref="float"/>.
+        /// </summary>
+        [Serializable]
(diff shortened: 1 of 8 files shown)


thestonefox:
This all looks fine to me, you can pull it out of being draft for a review if you want

fight4dream:
It is ready now

Runtime/Tracking/Velocity/VelocityEmitter.cs
```diff
@@ -20,9 +20,16 @@ public class VelocityEmitter : MonoBehaviour
         public class UnityEvent : UnityEvent<Vector3>
```
thestonefox: I wonder if this should be renamed to Vector3UnityEvent as there is now a FloatUnityEvent as well

fight4dream: Would that be a breaking change?

fight4dream: And i feel like that rename shall be a separate chore PR

thestonefox: No i don't think it would be a breaking change as it's still just internal to the class itself, changing the name of the internal unity class _shouldn't_ break anything elsewhere.

I don't think it needs to be a separate PR either. I dunno if @bddckr has time to have a view on this.

fight4dream: I can access the computer to check on this after few hours

bddckr: I'm not sure how Unity handles nested types in their serialization engine, so better check to see if it's a breaking change or not. :+1:

PR #485, opened 2020-01-02
fix(Tracking): cache tracker source when OnEnable - fixes #486

### LINKED ISSUE #486: Component Tracker Proxy does not cache the proxy source
version 21870267bd43c83fb05ed002ee18d55cfbb0d117

1. Create a game object and add ComponentTrackerProxy
2. Fill in its ProxySource with LeftControllerAlias
3. Change inspector to Debug mode and then Play

### Issue

Its CachedVelocityTracker is none. thus the component doesn't work

### Expected Behaviour

Its CachedVelocityTracker should be filled in with the velocity tracker on LeftControllerAlias.

fight4dream:
This fix will cache velocity tracker for this component to work
when the ProxySource is set up in editor time.

### DIFF
--- a/Runtime/Tracking/Velocity/ComponentTrackerProxy.cs
+++ b/Runtime/Tracking/Velocity/ComponentTrackerProxy.cs
@@ -5,6 +5,7 @@
     using Malimbe.MemberClearanceMethod;
     using Malimbe.XmlDocumentationAttribute;
     using Malimbe.PropertySerializationAttribute;
+    using Zinnia.Extension;
 
     /// <summary>
     /// Attempts to utilize the first <see cref="VelocityTracker"/> found on the given proxy <see cref="Component"/>.
@@ -22,6 +23,11 @@
         /// The cached <see cref="VelocityTracker"/> found on the proxy <see cref="Component"/>.
         /// </summary>
         protected VelocityTracker cachedVelocityTracker;
+
+        protected virtual void OnEnable()
+        {
+            SetCachedVelocityTracker();
+        }
 
         /// <inheritdoc />
         public override bool IsActive()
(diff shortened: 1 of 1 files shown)


fight4dream:
fixes #486

Runtime/Tracking/Velocity/ComponentTrackerProxy.cs
```diff
@@ -46,6 +51,14 @@ protected override Vector3 DoGetAngularVelocity()
         /// </summary>
         [CalledAfterChangeOf(nameof(ProxySource))]
         protected virtual void OnAfterProxySourceChange()
+        {
+            SetCachedVelocityTracker();
+        }
+
+        /// <summary>
+        /// Sets the VelocityTracker cache to the first found <see ref="VelocityTracker"> on the <see ref="ProxySource"> or any of its descendants.
```
bddckr:
```suggestion
        /// Sets <see ref="cachedVelocityTracker"> to the first found <see ref="VelocityTracker"> on the <see ref="ProxySource"> or any of its descendants.
```

Runtime/Tracking/Velocity/ComponentTrackerProxy.cs
```diff
@@ -46,6 +51,14 @@ protected override Vector3 DoGetAngularVelocity()
...
+        {
+            SetCachedVelocityTracker();
+        }
+
+        /// <summary>
+        /// Sets the VelocityTracker cache to the first found <see ref="VelocityTracker"> on the <see ref="ProxySource"> or any of its descendants.
+        /// </summary>
+        protected virtual void SetCachedVelocityTracker()
         {
             cachedVelocityTracker = ProxySource == null ? null : ProxySource.GetComponentInChildren<VelocityTracker>();
```
bddckr: We have a nice extension method for this:
```suggestion
            cachedVelocityTracker = ProxySource.TryGetComponent<VelocityTracker>(true);
```

Runtime/Tracking/Velocity/ComponentTrackerProxy.cs
fight4dream: done in new commit

Runtime/Tracking/Velocity/ComponentTrackerProxy.cs
fight4dream: done in new commit

PR #488, opened 2020-01-11
fix(Tracking): calculate properties correctly if dynamic destination fix #487

### LINKED ISSUE #487: TransformPropertyApplier does not honor the ApplyTransformations settings if IsTransitionDestinationDynamic set to true
### Steps to reproduce

1. use a model that you can see clearly its orientation
2. create an empty game object, add TransformPropertyApplier
3. set source = left controller alias, target = model
4. set IsTransitionDestinationDynamic = true
5. set ApplyTransformations to either Position only or Rotation Only

### Expected behavior

if set Position only, the model should not rotate
if set Rotation only, the model should not move

### Current behavior

the model will rotate and move

fight4dream:
This fix will calculate properties correctly if IsTransitionDestinationDynamic is true.
Before the fix, it does not honor the ApplyTransformations settings.

### DIFF
--- a/Runtime/Tracking/Modification/TransformPropertyApplier.cs
+++ b/Runtime/Tracking/Modification/TransformPropertyApplier.cs
@@ -399,9 +399,9 @@
 
                 if (dynamicDestination)
                 {
-                    destinationScale = Source.Scale;
-                    destinationRotation = Source.Rotation;
-                    destinationPosition = Source.Position;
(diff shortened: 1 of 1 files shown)


thestonefox:
I feel this should have a test to go with it to prove the problem and thus prove the fix

thestonefox:
Superseded by https://github.com/ExtendRealityLtd/Zinnia.Unity/pull/497

PR #495, opened 2020-04-13
Things that need doing - fixes #494 - fixes #385 - fixes #469

### LINKED ISSUE #385: GameObjectExtractors have inconsistent public APIs
The GameObjectExtractor inherited classes have differing public APIs on how to extract a custom component to a GameObject.

Some have a `Source` property which is required to be set and is then used in the inherited `Extract()` method

Whereas others don't have a `Source` property but create a new `Extract(source)` method.

e.g.
`TransformDataGameObjectExtractor` uses the `Source` property method

`PublisherContainerExtractor` uses the `Extract(source)` method

I propose that instead of picking one way or another, that all Extractors actually implement both ways.

This not only reduces the issue of breaking changes, but it also means the usage of an extractor is more customisable.

For example, the `Source` property method allows for a pre-defined source at edit time, whereas the `Extract(source)` method allows for runtime extraction without needing to set a `Source` property.

### LINKED ISSUE #469: Deprecate Vector2ComponentExtractor and Vector3ComponentExtractor
The `Vector2ComponentExtractor` and `Vector3ComponentExtractor` are doing the exact same job as the `Vector2ToFloat` and `Vector3ToFloat` components so they should be deprecated and removed.

It looks like only the `Vector2ComponentExtractor` is being used within the VRTK farmyard scene to create simulated pointer rotations with the mouse wheel so the impact of depreciation should be minimal.

### LINKED ISSUE #494: Data/Operation/Extraction/TransformDirectionExtractor UseLocal setting seems wrong way round
The `UseLocal` setting will return the global `Vector3` direction and not the direction local to the source

```
            switch (Direction)
            {
                case AxisDirection.Right:
                    return UseLocal ? Vector3.right : Source.transform.right;
                case AxisDirection.Up:
                    return UseLocal ? Vector3.up : Source.transform.up;
                case AxisDirection.Forward:
                    return UseLocal ? Vector3.forward : Source.transform.forward;
            }
```

This feels like it is the wrong way round.

`UseLocal` being true should return the local `Source` direction and if it's false it should return the global direction.

Changing this is technically a breaking change though

extendreality:


thestonefox:
@bddckr There's a lot here, most of it is just mundane tidying up. Don't feel like you need to give it any real thought :)

Runtime/Data/Operation/Extraction/ValueExtractor.cs
```diff
@@ -109,5 +109,21 @@ protected virtual void OnDisable()
...
         }
+
+        /// <summary>
+        /// Attempts to invoke the <see cref="Extracted"/> event if the <see cref="Result"/> is not null.
+        /// </summary>
+        /// <param name="data">The data to emit.</param>
+        /// <returns>Whether the event was invoked.</returns>
+        protected virtual bool InvokeEvent(TEventElement data)
+        {
+            if (Result == null)
```
bddckr, line 120:
```suggestion
            if (data == null)
```
Isn't this suggested change what is intended?

Runtime/Data/Operation/Cache/ValueCache.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <summary>
+        /// The backing field for <see cref="Data"/>.
+        /// </summary>
+        private TValue data;
+
+        /// <summary>
+        /// Clears the cache by setting it to the default value.
+        /// </summary>
+        [RequiresBehaviourState]
+        public virtual void ClearCache()
```
bddckr, line 55: Can't we use `[Cleared]` on the property `Data` instead? The only difference is that the (autogenerated) method will then be called `ClearData` instead - is that the reason why the attribute isn't used?

Or I guess it's due to wanting the potential ability to override this method?

Runtime/Data/Operation/Cache/ValueCache.cs
```diff
@@ -0,0 +1,68 @@
...
+            Data = default;
+        }
+
+        /// <summary>
+        /// Determines if the two given values are considered equal.
+        /// </summary>
+        /// <param name="a">The first value to check.</param>
+        /// <param name="b">The second value to check.</param>
+        /// <returns>Whether the two given values are considered equal.</returns>
+        protected abstract bool AreEqual(TValue a, TValue b);
```
bddckr, line 66: This could be virtual and offer a base implementation by returning `EqualityComparer<TValue>.Default.Equals(a, b)` I think. Worth it? As far as I can see this change would then remove the need to have subclasses implement this method for the `int`, `string` and `GameObject` cases at least (the `float` one wants the custom approximate check, and there might be others I didn't check).

Runtime/Data/Operation/Extraction/ValueExtractor.cs
thestonefox: the `data` maybe of a type that is not nullable (e.g. `Vector3`) but the `Result` will be of nullable type.

The reason for this is, the type can be nullable but the event payload can't be nullable because Unity can't display a nullable field like its non nullable counterpart

So a UnityEvent with Vector3 payload shows in the inspector properly, but a Vector3? payload doesn't show up in the inspector as you'd expect.

Runtime/Data/Operation/Cache/ValueCache.cs
thestonefox: Yeah probably could use Cleared, I'll add it and run the tests. I don't think there's any need to override the method, if you have a cache with a custom clear up then you can just write a custom method at that point i think

thestonefox: Putting `[Cleared]` on the `Data` property doesn't give me a `ClearData` method.

I would have thought it would, maybe it's because of my custom setter?

Runtime/Data/Operation/Extraction/ValueExtractor.cs
bddckr: 👍

Runtime/Data/Operation/Cache/ValueCache.cs
bddckr: Oh yeah might be. Let's just ignore this then!

PR #499, opened 2020-05-21
Feat/platform device associations

extendreality:


Runtime/Association/PlatformDeviceAssociation.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public string PlatformPattern { get; set; }
+        /// <summary>
+        /// A regular expression to match the name of the XR device that needs to be loaded.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public string XrSDKPattern { get; set; }
```
bddckr, line 25: `XrSdkPattern` follows the C# naming conventions a little bit better.

Runtime/Association/PlatformDeviceAssociation.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Holds <see cref="GameObject"/>s to (de)activate based on the current platform and loaded XR device type.
+    /// </summary>
+    public class PlatformDeviceAssociation : GameObjectsAssociation
+    {
+        /// <summary>
+        /// A regular expression to match the name of the current <see cref="RuntimePlatform"/>.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public string PlatformPattern { get; set; }
```
bddckr, line 19: `Application.platform` is using a known type, should this perhaps be a multi-select thing for picking enum values? Or is the issue making that work with the Unity inspector perhaps, and thus regex provides the same wanted outcome?

extendreality: A regular expression will also allow you to fuzzy check the platform, or do multiple platform checks, e.g. Android or Standalone

PR #503, opened 2020-06-06
Feat/adding stuff

extendreality:


Runtime/Process/Component/SourceTargetProcessor.cs
```diff
@@ -93,12 +101,24 @@ protected virtual void ApplySourcesToTargets(HeapAllocationFreeReadOnlyList<TSou
                     ApplySourceToTarget(currentSource, currentTarget);
                 }
 
+                if (!EqualityComparer<TSource>.Default.Equals(ActiveSource, currentSource))
+                {
+                    ActiveSourceChanged?.Invoke(currentSource);
```
thestonefox, line 106: I know this is emitted before the actual `ActiveSource` is changed...

But it does raise with the actual new ActiveSource and therefore technically you still have access to the original value because `ActiveSource` is still the old value during the event process.

I can make it so it emits with `ActiveSource` being changed (by adding some caching of values) but is this actually more useful to have the old value?

bddckr: Sure it's useful, but the docs on `ActiveSourceChanged` are wrong as is. Maybe just update them to include this detail?

thestonefox: What about if the event name is changed as well?

Because the event right now is past tense stating that the `ActiveSource` has changed.

What about making it `ActiveSoureChanging` and it is documented as being raised right before `ActiveSource` is updated.

bddckr: 👍 👍

PR #510, opened 2020-07-25
feat(Extraction): add extractors for ObjectFollower.EventData

fight4dream:
A collection of extractors that can extract specific data from
the ObjectFollower.EventData.

For example, an Interactable is using Follow Rigidbody.
When it is diverged, the event data can be feed into the
FollowerSourceExtractor to get the grabbing interactor.

### DIFF
--- a/Runtime/Tracking/Velocity/XRNodeVelocityEstimator.cs
+++ b/Runtime/Tracking/Velocity/XRNodeVelocityEstimator.cs
@@ -1,5 +1,6 @@
 ﻿namespace Zinnia.Tracking.Velocity
 {
+    using Malimbe.MemberClearanceMethod;
     using Malimbe.PropertySerializationAttribute;
     using Malimbe.XmlDocumentationAttribute;
     using System.Collections.Generic;
@@ -19,6 +20,13 @@
         public XRNode Node { get; set; } = XRNode.LeftHand;
 
         /// <summary>
+        /// An optional object to consider the source relative to when estimating the velocities.
+        /// </summary>
+        [Serialized, Cleared]
+        [field: DocumentedByXml]
+        public GameObject RelativeTo { get; set; }
+
+        /// <summary>
         /// A collection of node states.
         /// </summary>
         protected readonly List<XRNodeState> nodesStates = new List<XRNodeState>();
@@ -27,6 +35,7 @@
         protected override Vector3 DoGetVelocity()
         {
             GetNodeState().TryGetVelocity(out Vector3 result);
+            result = (RelativeTo != null ? RelativeTo.transform.rotation : Quaternion.identity) * result;
             return result;
         }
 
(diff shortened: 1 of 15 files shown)


fight4dream:
https://github.com/ExtendRealityLtd/Tilia.Mutators.ObjectFollower.Unity/issues/26

Runtime/Tracking/Follow/Operation/Extraction/FollowerSourceExtractor.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    using System;
+    using UnityEngine;
+    using UnityEngine.Events;
+    using Zinnia.Data.Operation.Extraction;
+
+    /// <summary>
+    /// Extracts the <see cref="GameObject"/> of the EventSource from a given <see cref="ObjectFollower.EventData"/>.
+    /// </summary>
+    public class FollowerSourceExtractor : GameObjectExtractor<ObjectFollower.EventData, FollowerSourceExtractor.UnityEvent>
```
thestonefox, line 11: Name needs to match the class type it is for:

so should be `ObjectFollowerEventDataSourceExtractor` (same for the others too)

fight4dream: Because i saw there is NotifierXXXExtractor which operates on a CollisionNotifier. So it is a hint for me to shorten the name. You still want the full name?

thestonefox: Yeah they should be the full name, that's the correct way of referring to them (older extractors may not adhere to this, but that's because they're old)

fight4dream:
ready for review again

Runtime/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -27,13 +35,16 @@ public class XRNodeVelocityEstimator : VelocityTracker
         protected override Vector3 DoGetVelocity()
         {
             GetNodeState().TryGetVelocity(out Vector3 result);
+            result = RelativeTo.transform.rotation * result;
```
thestonefox, line 38: This should be wrapped in a null check because `RelativeTo` is optional:

```
result = RelativeTo != null ? RelativeTo.transform.rotation * result : result;
```

Runtime/Tracking/Velocity/XRNodeVelocityEstimator.cs
```diff
@@ -27,13 +35,16 @@ public class XRNodeVelocityEstimator : VelocityTracker
...
             return result;
         }
 
         /// <inheritdoc />
         protected override Vector3 DoGetAngularVelocity()
         {
             GetNodeState().TryGetAngularVelocity(out Vector3 result);
+            GetNodeState().TryGetRotation(out Quaternion rotation);
+            result = RelativeTo.transform.rotation * rotation * result;
```
thestonefox, line 47: Again, wrap in null check:

```
result = (RelativeTo != null ? RelativeTo.transform.rotation : Quaternion.identity) * rotation * result;
```

should be acceptable I think.

fight4dream:
ready for review again

PR #515, opened 2020-08-23
Feats and fixes - fixes #514 - resolves #511

### LINKED ISSUE #511: from a float list, return the max/min/median etc.
## Is your feature request related to a problem?

in developing for handtracking,
i want the max of (middle, ring, pinky) pinch strength to be the grab axis

## Describe the solution you'd like

i think it is similar to the FloatAdder, only the inner operation change to 'max'
thus it takes a float observable list, and return the max of the elements

## Additional context

other operations such as Min, Median, Mode, Average

### LINKED ISSUE #514: Transform Direction Extractor UseLocal result not expected
### Environment

* Code source: github
* Code version: 18d34c1

### Steps to reproduce

Goal: make a up-right camera following HeadsetAlias

1. Set up tilia xr environment (unityxr, tracked alias, etc)
2. Create a camera named "Record Camera"
3. Create a game object "Follow Position", add component "TransformPositionExtractor"
4. Drag HeadsetAlias into Source field, Extracted event calls `Record Camera.transform.position`. We want the Record Camera follow the **world space position, so turn off UseLocal.**
image
5. Create a game object "Follow Rotation", add component "TransformDirectionExtractor"
6. Add the custom script AxesToRotation (see the Append). It takes two Vector3 axes and compute a Quaternion rotation.
7. Drag HeadsetAlias into Source field of direction extractor, Extracted event calls to AxesToRotation.Forward and AxesToRotation.EmitRotation(). We want the Record Camera's **world space forward direction, so turn off UseLocal.**
8. AxesToRotation rotation emitted event calls to `Record Camera.transform.rotation`. We want the upright orientation, so we use the ConstructFrom = Forward and Up vectors.
image
9. Add MomentProcess to both FollowPosition and FollowRotation, create a MomentProcessor to execute both of them.
image

### Current behavior

The Record Camera's position is followed correctly, but it is pointing in world space's z direction, not following the forward direction of HeadsetAlias.

For Transform Position Extractor, when UseLocal is off, it returns the source transform's **world space** position, while UseLocal is on, it returns the source transform's **local**Position
```
        protected override Vector3? ExtractValue()
        {
            ...
            return UseLocal ? Source.transform.localPosition : Source.transform.position;
```

However, the current Transform Direction Extractor, when UseLocal is off, it returns the Vector3.forward (which is independent of any source), while UseLocal is on, it returns the source transform's **world** space forward instead of **local** space forward
```
        protected override Vector3? ExtractValue()
        {
                ...
                case AxisDirection.Forward:
                    return UseLocal ? Source.transform.forward : Vector3.forward;
```

### Expected behavior

The transform direction extractor, when UseLocal is off should return the **source transform's world space** direction, and when UseLocal is on should return the **source transform's local space** direction .
i.e. `return UseLocal ? Source.transform.localRotation * Vector3.forward : Source.transform.forward;`

### Append

```
namespace VRToolkitExtras.Data.Type.Transformation
{
    using System;
    using Malimbe.PropertySerializationAttribute;
    using UnityEngine;
    using UnityEngine.Events;

    public class AxesToRotation : MonoBehaviour
    {
        [Serializable]
        public class QuaternionUnityEvent : UnityEvent<Quaternion>
        {
        }

        [Serialized]
        public Vector3 Forward { get; set; } = Vector3.forward;
        [Serialized]
        public Vector3 Up { get; set; } = Vector3.up;
        [Serialized]
        public Vector3 Right { get; set; } = Vector3.right;

        public enum ConstructFromType
        {
            ForwardAndUp,
            ForwardAndRight
        }
        [Serialized]
        public ConstructFromType ConstructFrom { get; set; } = ConstructFromType.ForwardAndUp;

        public Quaternion Rotation { get; private set; }

        public QuaternionUnityEvent RotationEmitted = new QuaternionUnityEvent();

        public void EmitRotation()
        {
            switch (ConstructFrom)
            {
                default:
                case ConstructFromType.ForwardAndUp:
                    {
                        Rotation = Quaternion.LookRotation(Forward, Up);
                        RotationEmitted?.Invoke(Rotation);
                    }
                    break;
                case ConstructFromType.ForwardAndRight:
                    {
                        Vector3 up = Vector3.Cross(Forward, Right);
                        Rotation = Quaternion.LookRotation(Forward, up);
                        RotationEmitted?.Invoke(Rotation);
                    }
                    break;
            }
        }
    }
}
```

thestonefox:


fight4dream (commented):
for the float aggregate functions,
What if the input array is empty? (count=0)
It seems most of them will throw errors.

thestonefox:
I'll add some checks

fight4dream:
If they will throw,

I think a solution is to have Failed event to the aggregator (just like the Failed event on extractor) has the flexibility to let end users handle the exception.

Another possible solution is to have a IsEmpty event on the observable list, but the user has to add more logic to check for empty first. If they forget to do that, the aggregator would still throw. 

So i prefer the first solution.

thestonefox:
I don't think it should have a failed event. It should aggregate to 0 I'd say.

If you sum up 0 records in SQL for example you get 0 you don't get told it had nothing.

The list you're aggregating on has events for checking the state of the list.

thestonefox:
Actually you don't get 0, you get `null` so yeah maybe an event determining that happened (not failed, because the operation didn't fail persay)

thestonefox:
ok i can't think of anything better than `failed`

fight4dream:
### Review on this zinnia branch on AngularTransformDrive

1. Setup Tracked alias, add Interactor in RightControllerAlias
image
2. Add AngularTransformDrive. Set its position=(0, 0.8, 0) euler=(-90, 0, 0), Drive Axis=Y axis, Drive Speed=90, DriveLimit = [-120, 120]
3. Go to its nested Interactable. Change its Cube position=(0, 0.016, 0.0105) Scale=(0.005, 0.01, 0.01). Also add a Cylinder Scale=(0.03, 0.02, 0.03)
image
image
4. Set the Interactable to use Follow Rotate Around Angular Velocity
image
5. Play and try to turn the knob. Can try to grab the knob at different angles.

### Expected Behaviour

The knob should turn and rotate only around its hinged axis.

### Current Behaviour

The knob is not turning around its hinged axis.
And grabbing the knob, turn it in other ways seem to make it move/rotate in unexpected ways

thestonefox:
try taking precision grab off

fight4dream:
Changed to Orientation Handle,
tested default (handle's Y pointing towards me),
It turns on its axle. but direction reversed (i want to turn it clockwise, it rotated anti-clockwise)
also not responsive.

then i tested rotate the handle = (0, 0, 180) (handle's Y pointing away from me),
same, direction reversed.

thestonefox:
Set grab offset to none. There's no need to have a grab offset

fight4dream:
same results. (Drive Y axis, Apply Y axis)

interestingly, when i rotate my controller by controller's y axis, the knob is more responsive.
so i think Apply Y is picking up the controller's Y axis.

Further testing with Drive Y, Apply **Z**
z axis is more align with controller's turning a knob action.
however, the knob rotates by its local Z axis
image

thestonefox:
The problem you have is your rotations are out of whack with the controller angular velocities of the controller.

If you want to rotate the controller around your wrist then that is the controller's z axis:

and because you've rotated your controllable object to fix the default rotation of the cylinder you then end up with a Y rotation as your forward.

so instead, don't rotate your controllable around the x90, and rotate your cylinder mesh around x90

Then your drive axis will still be on the Z matching the controller axis you're wanting to work with.

You can then rotate the controllable however you want as its z is still matching up

rotations

thestonefox:
(also it's worth increasing the Source Multiplier on the RotateAroundAngularVelocity component to 1,1,1 so it tracks better to a smaller object)

fight4dream:
It is the Drive's axis setting that gives us freedom in the design choice of how we want to orient our model inside.
and very importantly, that axis is local space to the Drive.
thus it is by choice that one would want a working knob that rotates on the drive's y axis.

Notice I had not change the orientation of these two transforms (Interactable Container, Interactions.Interactable)
because they are delibrately controlled by the drive facade.
image

the knob meshes (cylinder and the cube), just like other interactables, are purely cosmetic. it happens that i orient it so that its rotation axis aligns with that of the drive. it shouldn't matter the mesh's orientation at all. In any case, it is the Drive's Axis that should be honored. and the cosmetic just show more clearly if the result is correct or not.

Also, remember if we use the original rotate method (FollowTransformPositionDifferenceRotate), drive axis Y rotates as expected.
Hence, it is expected to work whether we choose the drive axis to be x,y or z.

PR #520, opened 2020-11-24
Feat/tracking feats

fight4dream:
feat(Tracking): add new velocity trackers
feat(Tracking): extract as IProcessable for two components


Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// Determines whether to extract the local property or the world property.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public bool UseLocal { get; set; }
+
+        /// <inheritdoc />
+        protected override Vector3 DoGetAngularVelocity()
+        {
+            return UseLocal ? transform.localRotation * AngularVelocity : AngularVelocity;
```
thestonefox, line 36: isn't this check pointless? why even have the UseLocal on this type?

Runtime/Tracking/Velocity/RigidbodyVelocityTracker.cs
```diff
@@ -0,0 +1,46 @@
...
+        [Serialized, Cleared]
+        [field: DocumentedByXml]
+        public Rigidbody Source { get; set; }
+
+        /// <summary>
+        /// The velocity to return.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public Vector3 Velocity { get; set; }
```
thestonefox, line 25: why can you specify a velocity (and angular velocity)? these don't seem to be used anywhere, only Source is used.

thestonefox (changes requested):

Runtime/Tracking/Velocity/ArtificialVelocityApplier.cs
```diff
@@ -1,189 +1,55 @@
 ﻿namespace Zinnia.Tracking.Velocity
```
thestonefox, line 1: Why is this being extracted? This class doesn't use an Update method to process anything it simply calls Apply. So if anything the wrapper should be the other way round no?

This should stay as is, and if you want something that is iProcessable then you just have it call the Apply method in this class on every moment?

thestonefox (changes requested):

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace Zinnia.Tracking.Velocity
```
thestonefox, line 1: what's the use case for this class?

thestonefox (changes requested):

Runtime/Tracking/Velocity/RigidbodyVelocityTracker.cs
fight4dream: Ah i forgot to remove those properties after dup from constant velocity tracker
It should only have the Source property

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
fight4dream: My move in place uses velocity trackers as speed input. To debug an issue, i use this class to isolate the cause and find out average velocity tracker is part of the problem (frame rate dependant)

Other than debug purpose, this class is for “velocity tracker” whereas “Constant Force” component is for rigidbody. For example, for simulator/non vr mode, I had to simulate a grab and throw. Without this class, we had to key an animation for the interactor to release the object at certain angle and speed. It is a hard way. With this class, we create a constant tracker parented to playAreaAlias, with UseLocal=true, Velocity=(0,0.5,0.5), the object will be thrown consistently to direction desired and speed desired.

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
fight4dream: It is analogous to the Constant Force component’s world space force or relative force. 
Here we specify whether the velocity is world space or relative to self 
Also see above simulated throw use case.

Runtime/Tracking/Velocity/ArtificialVelocityApplier.cs
fight4dream: Notice originally when call Apply() once, it starts a coroutine to “process” the deceleration every frame (coroutine works post Update() but before LateUpdate()). So it is NOT equivalent to call Apply() in the moment we want to.

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
thestonefox: ok. this class needs a complimenting test to go along with it

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
thestonefox: sorry, I read the line wrong, I didn't see the `transform.localRotation *`

this also needs a test for the class

Runtime/Tracking/Velocity/RigidbodyVelocityTracker.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace Zinnia.Tracking.Velocity
+{
```
thestonefox, line 2: class needs a test

thestonefox (changes requested):

Runtime/Tracking/Velocity/ArtificialVelocityApplier.cs
thestonefox: right I see, so its the Deceleration is the thing that is being moved into the Process, I understand now

thestonefox, line 1: Do the tests still pass for these updates?

fight4dream: image
yes they are.

Runtime/Tracking/Velocity/RigidbodyVelocityTracker.cs
fight4dream: image
will force push test class when all done

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
fight4dream: image

Runtime/Tracking/Velocity/RigidbodyVelocityTracker.cs
fight4dream: removed

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
fight4dream: done

fight4dream:
image

fight4dream:
@thestonefox is this what you wanted?

thestonefox:
> @thestonefox is this what you wanted?

No, I mean just delete that last commit, I'll do a commit that fixes that stuff

fight4dream:
done~

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <inheritdoc />
+        protected override Vector3 DoGetAngularVelocity()
+        {
+            return UseLocal ? transform.localRotation * AngularVelocity : AngularVelocity;
+        }
+
+        /// <inheritdoc />
+        protected override Vector3 DoGetVelocity()
+        {
+            return UseLocal ? transform.localRotation * Velocity : Velocity;
```
thestonefox, line 42: why is this multiplying by local rotation for velocity?:

thestonefox (changes requested):

Runtime/Tracking/Velocity/ConstantVelocityTracker.cs
fight4dream: https://github.com/ExtendRealityLtd/Zinnia.Unity/pull/520#discussion_r530399569

fight4dream: https://github.com/ExtendRealityLtd/Zinnia.Unity/pull/520#discussion_r530400950

PR #532, opened 2021-03-19
Update RotateAroundAngularVelocity.cs

hawkwood:
When using an angular transform drive as a handleless door knob, this class is used to track the interactor's "wrist" rotation, but as it is now the this only works when drive/interactor are aligned with world space.  By using this new option, the angular velocity will be transformed to the target's space, and thus applied properly.

### DIFF
--- a/Runtime/Tracking/Follow/Modifier/Property/Rotation/RotateAroundAngularVelocity.cs
+++ b/Runtime/Tracking/Follow/Modifier/Property/Rotation/RotateAroundAngularVelocity.cs
@@ -30,6 +30,13 @@
         [Serialized]
         [field: DocumentedByXml]
         public Vector3State ApplyToAxis { get; set; }
+        
+        /// <summary>
+        /// When true, transforms the angular velocity to be in target's space instead of world space.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public bool InTargetSpace { get; set; }
 
         /// <summary>
         /// Sets the <see cref="SourceMultiplier"/> x value.
@@ -98,7 +105,10 @@
                 return;
             }
 
-            Vector3 input = AngularVelocitySource.GetAngularVelocity();
+            Vector3 input = InTargetSpace
+                ? target.transform.parent.InverseTransformVector(AngularVelocitySource.GetAngularVelocity())
+                : AngularVelocitySource.GetAngularVelocity();
+                            
             input.Scale(SourceMultiplier);
             input.Scale(ApplyToAxis.ToVector3());
 


thestonefox:
If this is fixing a bug then please raise a bug report to go along with it so it can be shown what it is fixing

Runtime/Tracking/Follow/Modifier/Property/Rotation/RotateAroundAngularVelocity.cs
```diff
@@ -30,6 +30,13 @@ public class RotateAroundAngularVelocity : PropertyModifier
         [Serialized]
         [field: DocumentedByXml]
         public Vector3State ApplyToAxis { get; set; }
+        
+        /// <summary>
+        /// When true, transforms the angular velocity to be in target's space instead of world space.
+        /// </summary>
+        [Serialized]
+        [field: DocumentedByXml]
+        public bool InTargetSpace { get; set; }
```
thestonefox, line 39: Nice solution to prevent any regression issues, but I think this should be defaulted to `true` and any regressions can just have this set to `false` so just change this to:

`public bool InTargetSpace { get; set; } = true;`

Runtime/Tracking/Follow/Modifier/Property/Rotation/RotateAroundAngularVelocity.cs
```diff
@@ -99,6 +106,12 @@ protected override void DoModify(GameObject source, GameObject target, GameObjec
             }
 
             Vector3 input = AngularVelocitySource.GetAngularVelocity();
+            
+            if (InTargetSpace)
```
thestonefox, line 110: Can combine this into a single line for neatness:

```
Vector3 input = InTargetSpace
    ? target.transform.parent.InverseTransformVector(AngularVelocitySource.GetAngularVelocity())
    : AngularVelocitySource.GetAngularVelocity();
```

thestonefox (changes requested):
You also need to change your commit message (via `git commit --amend`)

your subject line does not meet the contribution requirements and your subject body has no character breaks and so the lines are too long, it should read something like this:

```
feat(Tracking): allow local space rotation around angular velocity

When using an angular transform drive as a handleless door knob, 
this class is used to track the interactor's "wrist" rotation,
but as it is now the this only works when drive/interactor are
aligned with world space.  By using this new option, the angular
velocity will be transformed to the target's space, and thus
applied properly.
```

Runtime/Tracking/Follow/Modifier/Property/Rotation/RotateAroundAngularVelocity.cs
hawkwood: I disagree that this should be "true" by default, because it does notably change the output of the function. If someone is using this script elsewhere, this little change could severely impact them, whereas for the purposes of using the angular transform drive as a twistable object, there will already be additional setup steps, so adding one more of setting this value to true is trivial.

thestonefox: yeah ok, i agree with this

Runtime/Tracking/Follow/Modifier/Property/Rotation/RotateAroundAngularVelocity.cs
hawkwood: done.

thestonefox:
All this needs squashing in to one commit, with the commit message body not exceeding 72 characters per line.

`git rebase -i HEAD~3` to start squashing

hawkwood:
I messed up my branch trying to rebase and squash, so I made a new branch and am creating a new pull request.

PR #552, opened 2022-01-04
enable non xr use for zinnia

fight4dream:
not to be merged directly
--------------
both XRNodeHapticPulser and XRNodeVelocityEstimator uses XRNode for serialized property,
cannot be modified readily to ignore those properties without losing the serialized data.
I chose to complete #if out the components.

PlatformDeviceAssociation (and deprecated LoadedXrDeviceAssociation) uses XRNode, XRDevice, XRSettings
but some of the logic can be ignored, so I chose to #if just those parts.
-------------------------------------
Zinnia.Runtime.asmdef added versionDefines
XRNode and InputDevices are within com.unity.modules.xr
XRDevice and XRSettings are within com.unity.modules.vr
com.unity.modules.vr depends on com.unity.modules.xr

although technically we would expect them both to be enabled in a vr project,
I chose to make separate define symbols for each of them.
--------------------------------
The following Tilia have reference to XR/VR:
"io.extendreality.tilia.camerarigs.spatialsimulator.unity": "1.4.21",
"io.extendreality.tilia.camerarigs.unityxr": "1.5.9",
"io.extendreality.tilia.camerarigs.xrpluginframework.unity": "1.1.10",
"io.extendreality.tilia.output.interactorhaptics.unity": "1.1.34",
"io.extendreality.tilia.sdk.oculusintegration.unity": "1.3.19",
"io.extendreality.tilia.sdk.steamvr.unity": "1.1.22",

### DIFF
--- a/Runtime/Association/LoadedXrDeviceAssociation.cs
+++ b/Runtime/Association/LoadedXrDeviceAssociation.cs
@@ -5,7 +5,9 @@
     using System;
     using System.Text.RegularExpressions;
     using UnityEngine;
+#if UNITY_XR || UNITY_VR
     using UnityEngine.XR;
+#endif
 
     /// <summary>
     /// Holds <see cref="GameObject"/>s to (de)activate based on the loaded XR device's name.
@@ -23,7 +25,11 @@
         /// <inheritdoc/>
         public override bool ShouldBeActive()
         {
+#if UNITY_VR
             return Regex.IsMatch(XRSettings.loadedDeviceName, XrDeviceNamePattern);
+#else
+            return false;
+#endif
         }
     }
 }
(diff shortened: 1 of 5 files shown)


fight4dream:
@thestonefox these are the basic changes to be made. I am not going to make this pull official as there may be other workarounds or things to be done differently (to support spatial simulator, and interactor haptics (xbox controller for example))

thestonefox:
looking over this, these ifdefs basically render a bunch of classes just empty.

What's the benefit of this? these classes can be present in a non-vr app and cause no issue, right?

thestonefox:
am i missing something about the serialization of XRNode causing an issue?

fight4dream:
@thestonefox non-vr apps would not have imported com.unity.modules.xr and com.unity.modules.vr packages
and a large part of zinnia/tilia do indeed does not depend on those packages.

without those packages, it would have been compile errors (and thus also serialization issue with XRNode for example)

moreover, the idea that a spatial simulator absolutely independent of those packages could take zinnia and some tilia a competitive candidate for more audience.

Issue #557, opened 2022-01-31
On Unpause (5/5 runs) - StraightLineCast has 0 length array


studentutu:
### Environment
Unity 2020.3.16f1

* Code source: Zinnia 
* Code version: 1.40
* Platform version: Android 10 ( API 29) , Pico VR ( old/new Unity SDK)
* 3rd party dependencies: None

### Steps to reproduce

1) Create simple rig with CameraRigs.TrackedAlias and 2 interactable pointers
2) Add Straight Pointers for left/right hand. Make it always visible. Set activations so that the activation event is always processed (pointer is always active)
4) Build Android Project and run on Pico
5) Launch App
6) Press Home button to navigate to Pico Quick System Menu. Wait for few seconds.
6) Go back to your app ( continue your app)
7) Sometimes it will throw out of boundary exceptions. ( 5/5) tries

This is also reproducible on Quest 1.

### Expected behavior

It should have a check for length before using indecies.

### Current behavior

ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
  ?, line 0, in System.ThrowHelper.ThrowArgumentOutOfRangeException
  ?, line 0, in System.ThrowHelper.ThrowArgumentOutOfRangeException
  ?, line 0, in System.Collections.Generic.List`1[T].set_Item
  ?, line 0, in Zinnia.Cast.StraightLineCast.GeneratePoints
  ?, line 0, in Zinnia.Cast.StraightLineCast.DoCastPoints
  ?, line 0, in Zinnia.Process.Moment.MomentProcess.ProcessNow
  ?, line 0, in Zinnia.Process.Moment.MomentProcessor.Process
  ?, line 0, in UnityEngine.Camera+CameraCallback.Invoke

Briefly describe what is actually happening.
![image](https://user-images.githubusercontent.com/18601652/151765081-0c550900-ffca-44ea-a64d-93fde99a9ec2.png)

studentutu:
as well as this issue : 

```
NullReferenceException: Object reference not set to an instance of an object
  File "<03c4cef74c314705b9055b51c799ab32>", line 0, in Zinnia.Cast.PointsCast.OnAfterTargetHitChange
  File "<03c4cef74c314705b9055b51c799ab32>", line 0, in Zinnia.Cast.PointsCast.OnEnable
[...]

PR #558, opened 2022-02-04
fix(StraightLineCast) : fixed issue #557

studentutu:
fix(StraightLineCast) : fixed issue when PointsCast is disabled, but state is still valid. Issue #557

Also, refactored part for points array to be of a proper size before using them both in PointsCast and in ParabolicLineCast

### DIFF
--- a/Runtime/Cast/FixedLineCast.cs
+++ b/Runtime/Cast/FixedLineCast.cs
@@ -35,9 +35,11 @@ namespace Zinnia.Cast
         {
             Vector3 originPosition = Origin.transform.position;
             Vector3 destinationPosition = originPosition + Origin.transform.forward * CurrentLength;
-
-            points[0] = originPosition;
-            points[1] = DestinationPointOverride != null ? (Vector3)DestinationPointOverride : destinationPosition;
+            if (points.Count > 1)
+            {
+                points[0] = originPosition;
+                points[1] = DestinationPointOverride != null ? (Vector3)DestinationPointOverride : destinationPosition;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Runtime/Cast/FixedLineCast.cs
```diff
@@ -35,9 +35,11 @@ protected override void GeneratePoints()
         {
             Vector3 originPosition = Origin.transform.position;
             Vector3 destinationPosition = originPosition + Origin.transform.forward * CurrentLength;
-
-            points[0] = originPosition;
-            points[1] = DestinationPointOverride != null ? (Vector3)DestinationPointOverride : destinationPosition;
+            if (points > 1)
```
thestonefox, line 38: this isn't valid code, if you're trying to check for the length of the points List then you need to use `points.Count`

so:

```
if (points.Count > 1)
```

Runtime/Cast/ParabolicLineCast.cs
```diff
@@ -206,16 +206,19 @@ protected virtual void GeneratePoints(Vector3 forward, Vector3 down)
...
-
-            curvePoints[0] = Origin.transform.position;
-            curvePoints[1] = forward + (Vector3.up * CurveOffset);
-            curvePoints[2] = down;
-            curvePoints[3] = down;
-
             points.Clear();
-            foreach (Vector3 generatedPoint in BezierCurveGenerator.GeneratePoints(SegmentCount, curvePoints))
+
+            if (curvePoints > 3)
```
thestonefox, line 211: same here, not valid code, you need to use `curvePoints.Count`

thestonefox (changes requested):
The commit message also needs fixing.

Commit messages should not be in past tense (e.g. fixed) and should be no longer than 72 characters per line, nor should they link to github issues.

thestonefox:
This pull request is a bit all over the place now with extra commits and unnecessary merges.

Rather than keep working on fixing this, i'll close it and i'll just add the changes required to a pull request I'm doing:

https://github.com/ExtendRealityLtd/Zinnia.Unity/pull/560

commit: https://github.com/ExtendRealityLtd/Zinnia.Unity/pull/560/commits/c4444e6e03ed7ba0d5a8b504ea15b584c3454a60

PR #571, opened 2022-06-17
fix(Pointer): pointer elements null checks

ArtemkaKun:
Null checks for pointer elements (Origin, RepeatedElement, Destination) were implemented because of nulls when pointers are not used (for example in networking).

### DIFF
--- a/Runtime/Pointer/ObjectPointer.cs
+++ b/Runtime/Pointer/ObjectPointer.cs
@@ -331,6 +331,11 @@
         {
             get
             {
+                if (ValidatePointerElements() == false)
+                {
+                    return false;
+                }
+                
                 if (!enabled)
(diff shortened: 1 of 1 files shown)


thestonefox:
Can you raise a relevant issue/bug to go along with the pull request so it can be demonstrated to what it is resolving

ArtemkaKun:
Sure, give me a sec

ArtemkaKun:
Hmm, while creating an issue about what this change actually fixes, I found the actual problem that causes it - for some reason when you have the PointerElements object in the ObjectPointer.Straight prefab variant deactivated, in a build, null errors appear from the ObjectPointer class.

image

So I will close this PR since a source of error is not in the code.

PR #573, opened 2022-08-09
fix(CameraOverlay): change scaled time to unscaled time

SamuelACook:
The deltaTime and WaitForSeconds get affected by the time scale
which can cause the overlay to never fade in or out when time scale
is set to approximately 0. Switching to unscaledDeltaTime and
WaitForSecondsRealtime will always fade in and out no matter
the time scale.

### DIFF
--- a/Runtime/Visual/CameraColorOverlay.cs
+++ b/Runtime/Visual/CameraColorOverlay.cs
@@ -248,7 +248,7 @@
         /// <summary>
         /// Delays the reset of <see cref="blinkRoutine"/> by <see cref="AddDuration"/> plus <see cref="AppliedDuration"/> seconds.
         /// </summary>
-        protected WaitForSeconds resetDelayYieldInstruction;
+        protected WaitForSecondsRealtime resetDelayYieldInstruction;
         /// <summary>
         /// The event data to be emitted throughout the process.
         /// </summary>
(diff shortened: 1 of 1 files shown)


thestonefox:
This seems fine, but I'm wondering if it should be a public property bool to use deltaTime or unscaledDeltaTime, because technically this will cause a breaking change albeit that it is seen as a bug?

Is there any reason it could be deltaTime and therefore people already using this expect it to be deltaTime and this change could cause a break to their code?

If so then we should probably introduce a new bool like `public bool UseUnscaledTime {}` which is false by default so no breaking changes are included.

What do you think? if you think there should never be a case using Delta time then that's fine, we could also retro fix it later on if anyone complains of a break.

SamuelACook:
I think for CameraColorOverlay its functionality use case would never require Delta time. I think this because we have properties to modify the duration of the blink, and modifying the time scale would essentially be doing that but forcing it. While time scale is used for many aspects of gameplay, blinks should always be representative of the duration properties. If someone is using time scale to manipulate the duration of the blink they can achieve this effect by modifying the duration properties based on the time scale. This change should not break functionality but just adjust the duration of the blink if time scale is being used.

I made appended changes to the commit that I missed in the first pull request. I had to update the WaitForSeconds to WaitForSecondsRealtime.

thestonefox (approved):

